PR #30, opened 2019-01-26
Add support for Connection object.

rhurey:
Expose internalData object on base Recognizer class. This is used to allow the fromRecognizer* pattern on the connection object and other future objects that will be conjured from a recognizer.
Seperate the concept of a connection being established from one being configured with context information via the speech.config message.
Support multiple recognize* iterations being done in the same connection. Previously each recognizeOnce / startContinuousRecog call established a new connection.

Add & Update Tests.

### DIFF
--- a/src/common.speech/IntentServiceRecognizer.ts
+++ b/src/common.speech/IntentServiceRecognizer.ts
@@ -19,7 +19,6 @@ import {
     CancellationErrorCodePropertyName,
     EnumTranslation,
     IntentResponse,
-    RequestSession,
     ServiceRecognizerBase,
     SimpleSpeechPhrase,
     SpeechHypothesis,
@@ -42,22 +41,20 @@ export class IntentServiceRecognizer extends ServiceRecognizerBase {
         connectionFactory: IConnectionFactory,
         audioSource: IAudioSource,
         recognizerConfig: RecognizerConfig,
-        recognizer: IntentRecognizer,
-        intentDataSent: boolean) {
+        recognizer: IntentRecognizer) {
         super(authentication, connectionFactory, audioSource, recognizerConfig, recognizer);
         this.privIntentRecognizer = recognizer;
-        this.privIntentDataSent = intentDataSent;
+        this.privIntentDataSent = false;
     }
 
     public setIntents(addedIntents: { [id: string]: AddedLmIntent; }, umbrellaIntent: AddedLmIntent): void {
         this.privAddedLmIntents = addedIntents;
         this.privUmbrellaIntent = umbrellaIntent;
+        this.privIntentDataSent = true;
     }
 
     protected processTypeSpecificMessages(
         connectionMessage: SpeechConnectionMessage,
-        requestSession: RequestSession,
-        connection: IConnection,
         successCallback?: (e: IntentRecognitionResult) => void,
         errorCallBack?: (e: string) => void): void {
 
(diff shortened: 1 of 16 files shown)


src/common.speech/IntentServiceRecognizer.ts
```diff
@@ -42,22 +41,20 @@ export class IntentServiceRecognizer extends ServiceRecognizerBase {
...
         super(authentication, connectionFactory, audioSource, recognizerConfig, recognizer);
         this.privIntentRecognizer = recognizer;
-        this.privIntentDataSent = intentDataSent;
+        this.privIntentDataSent = false;
     }
 
     public setIntents(addedIntents: { [id: string]: AddedLmIntent; }, umbrellaIntent: AddedLmIntent): void {
         this.privAddedLmIntents = addedIntents;
         this.privUmbrellaIntent = umbrellaIntent;
+        this.privIntentDataSent = true;
```
fmegen: >this.privIntentDataSent = true; [](start = 8, length = 31)

why is this set to true?

rhurey: The goal behind the variable was to see if the caller had added any intents to the recognizer or not.

src/sdk/Recognizer.ts
```diff
@@ -87,6 +93,15 @@ export abstract class Recognizer {
         this.dispose(true);
     }
 
+    /**
+     * @Internal
+     * Internal data member to support fromRecognizer* pattern methods on other classes.
+     * Do not use externally, object returned will change without warning or notive.
+     */
+    public get internalData(): object {
```
fmegen: > public get internalData(): [](start = 3, length = 27)

can we hide this here or otherwise this will end up in the public api for the recognizer?

src/sdk/TranslationRecognizer.ts
```diff
@@ -184,23 +183,17 @@ export class TranslationRecognizer extends Recognizer {
...
-            this.privReco = this.implRecognizerSetup(
-                RecognitionMode.Conversation,
-                this.properties,
-                this.audioConfig,
-                new TranslationConnectionFactory());
+            this.implRecognizerStop();
 
             this.implRecognizerStart(
-                this.privReco,
+                RecognitionMode.Conversation,
```
fmegen: > RecognitionMode.Conversation, [](start = 15, length = 30)

can this be overridden by properties? (which means we should take it from the properties instead of hardcoding)

tests/Settings.ts
```diff
@@ -14,8 +14,8 @@ export class Settings {
     // Endpoint should reduce standard speech timeout to value specified in SpeechServiceTimeoutSeconds
     // If undefined, production timeout of 10 seconds will be used, but at the cost of greatly incrased test
     // duration.
-    public static SpeechTimeoutEndpoint: string;
-    public static SpeechTimeoutKey: string;
+    public static SpeechTimeoutEndpoint: string = "wss://fedev-reconnect.develop2.cris.ai/speech/recognition/conversation/cognitiveservices/v1?language=en-us";
+    public static SpeechTimeoutKey: string = "c42994717bb44e9b843188ab0ad2d7b6";
```
fmegen: > public static SpeechTimeoutKey: string = "c42994717bb44e9b843188ab0ad2d7b6"; [](start = 3, length = 77)

(just curious) do we want this kind of information in the public repo?

rhurey: Gah! No.

rhurey: Keys were regenerated.

src/sdk/Recognizer.ts
rhurey: Given our mixed pattern of factory and constructor I didn't find a way to support all the FromRecognizer* methods in plan and not expose some data that was public that wasn't more invasive.

PR #40, opened 2019-03-13
Add proxy support for nodeJS.

rhurey:
As the proxy package depends on node features, when the webpack is built this change will skip building it in and will replace any use of it with an exception. Since the use is conditional to using a WebSocket library that also only works in node, this should be safe.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,15 +37,17 @@
     "@types/jest": "^23.3.10",
     "@types/node": "^10.7.0",
     "@types/request": "^2.48.1",
+    "@types/ws": "^6.0.1",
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -74,7 +76,7 @@
     "usePathForSuiteName": "true"
   },
   "dependencies": {
-    "@types/ws": "^6.0.1",
+    "https-proxy-agent": "^2.2.1",
```
fmegen: >  "https-proxy-agent": "^2.2.1" [](start = 2, length = 31)

do we need a legal review and add this to our open-source dependencies?

PR #41, opened 2019-03-15
Improve telemetry sent to Speech Service.

rhurey:
Send user latency telemetry on time from SDK acquisition of audio frames until first hypothesis returned and phrase returned.
Add Telemetry on type of audio source and label for Microphone if available.

Add more telemetry tests.
Add specific test for replay buffer.

### DIFF
--- a/src/common/Stream.ts
+++ b/src/common/Stream.ts
@@ -11,6 +11,7 @@ import { IStreamChunk } from "./Stream";
 export interface IStreamChunk<TBuffer> {
     isEnd: boolean;
     buffer: TBuffer;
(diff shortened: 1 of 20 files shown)


src/common/Stream.ts
```diff
@@ -11,6 +11,7 @@ import { IStreamChunk } from "./Stream";
 export interface IStreamChunk<TBuffer> {
     isEnd: boolean;
     buffer: TBuffer;
+    timeRecieved: number;
```
fmegen: >cie [](start = 10, length = 3)

typo

PR #45, opened 2019-03-27
Update Stream code & Add AudioWorklet support

rhurey:
Add AudioWorklet processing capability to move (basic) audio capture to use the audio worklet in browsers that support it. PCM recording work is still done on the JScript thread.

Add shell AudioWorklet to proxy audio frames over.
Add get / set property methods to AudioConfig

Add stream capability to accumulate audio into 4K frame sizes before sending to server. (TODO: This should be made user configurable), also applies to microphone input.


src/common.browser/MicAudioSource.ts
```diff
@@ -208,6 +218,12 @@ export class MicAudioSource implements IAudioSource {
         });
     }
 
+    public setProperty(name: string, value: string): void {
+        if (name === AudioWorkletSourceURLPropertyName) {
+            this.privRecorder.setWorkletUrl(value);
+        }
```
fmegen: how about raising an exception in case the property is not recognized

src/sdk/Audio/AudioConfig.ts
```diff
@@ -178,6 +201,22 @@ export class AudioConfigImpl extends AudioConfig implements IAudioSource {
         return this.privSource.events;
     }
 
+    public setProperty(name: string, value: string): void {
+        Contracts.throwIfNull(value, "value");
+
+        if (undefined !== this.privSource.setProperty) {
+            this.privSource.setProperty(name, value);
+        }
```
fmegen: how about raising an exception in case no priv property bag is defined

rssfrncs:
Is there a reason PCM recording can't be moved to an audio worklet? We see the following deprecation warning in the console.

```
[Deprecation] The ScriptProcessorNode is deprecated. Use AudioWorkletNode instead. (https://bit.ly/audio-worklet)

PR #49, opened 2019-04-04
Prevent webpack from loading https-proxy-agent.

### LINKED ISSUE #48: Cannot webpack v.1.4.0 due to https-proxy-agent issue.
Despite the effort to exclude <http-proxy-agent> from browser package in #40 it is still referenced by the main entry point in package.json. Which results in a build error in a consuming application.

Domratchev:
Fixes #48.
Another possible fix would be to use:
  "browser": "distrib/browser/microsoft.cognitiveservices.speech.sdk.bundle.js",

Removal of renameHttpsAgent function from `gulpfile.js` should also be considered if the proposed solution is accepted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,9 @@
     "type": "git",
     "url": "https://github.com/Microsoft/cognitive-services-speech-sdk-js"
   },
+  "browser": {
+    "https-proxy-agent": false
+  },
   "main": "distrib/lib/src/sdk/Exports.js",
   "types": "distrib/lib/src/sdk/Exports.d.ts",
   "files": [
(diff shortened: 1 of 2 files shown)


rhurey:
The gulpfile should be cleaned up also. 

Thanks for catching this, and for the fix.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "microsoft-cognitiveservices-speech-sdk",
   "author": "Microsoft Corporation",
   "homepage": "https://docs.microsoft.com/azure/cognitive-services/speech-service/",
-  "version": "1.3.0-alpha.0.1",
+  "version": "1.4.1",
```
mahilleb-msft: Please leave version as is, version is fixed up from our release branches:

```suggestion
  "version": "1.3.0-alpha.0.1",
```

Domratchev: Done. See the latest (squashed) commit.
You should also consider dropping `gulp-change` from devDependencies, as it is not used.

mahilleb-msft: thanks!

PR #50, opened 2019-04-05
Add demo bundle project

rhurey:
Add a test to ensure that developers consuming the npm package can webpack successfully.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -60,6 +60,16 @@ jobs:
     displayName: Publish test results
     condition: eq(variables['SPEECHSDK_RUN_TESTS'], 'true')
   - bash: |
+        cd tests/packaging
+        npm install
+        npm install ../../microsoft-cognitiveservices-speech-sdk-$SPEECHSDK_SEMVER2NOMETA.tgz
+        npm run bundle
+        E=$?
+        cd ../..
+        exit $E
+    displayName: Run Test Bundle
+    condition: eq(variables['SPEECHSDK_RUN_TESTS'], 'true')
+  - bash: |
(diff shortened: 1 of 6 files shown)


ci/build.yml
```diff
@@ -59,6 +59,16 @@ jobs:
   - task: PublishTestResults@2
     displayName: Publish test results
     condition: eq(variables['SPEECHSDK_RUN_TESTS'], 'true')
+  - bash: |
```
mahilleb-msft: Please either
- add `set -e` at the beginning to set that it stops on first failure (non-zero exit code) or
- chain the commands such as here
  ```
  a &&
  b &&
  c
  ```

ci/build.yml
```diff
@@ -59,6 +59,16 @@ jobs:
...
+        cd tests/packaging
+        npm install
+        npm install ../../microsoft-cognitiveservices-speech-sdk-$SPEECHSDK_SEMVER2NOMETA.tgz
+        npm run bundle
+        E=$?
```
mahilleb-msft: The cd at the end isn't necessary; once the step (process) exits, current directory is gone.
Hence, no need to store the exit code and return it either.
-> drop the last three lines

ci/build.yml
```diff
@@ -59,6 +59,16 @@ jobs:
   - task: PublishTestResults@2
     displayName: Publish test results
     condition: eq(variables['SPEECHSDK_RUN_TESTS'], 'true')
+  - bash: |
+        cd tests/packaging
+        npm install
```
mahilleb-msft: please use `npm ci` instead for restoring in a CI setting

PR #52, opened 2019-04-05
Provide es2015 module.

Domratchev:
Migrate from uglify to terser.

Observe dramatic reduction in minified bundle size (about 25%).
Even better results with webpack from npm package due to enabled tree shaking.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,25 +1,28 @@
 (function () {
   'use strict';
-  var gulp = require("gulp");
+  var gulp = require('gulp');
   var ts = require('gulp-typescript');
   var sourcemaps = require('gulp-sourcemaps');
-  var tslint = require("gulp-tslint");
-  var uglify = require('gulp-uglify');
+  var tslint = require('gulp-tslint');
+  var terser = require('gulp-terser');
   var rename = require('gulp-rename');
   var pump = require('pump');
   var webpack = require('webpack-stream');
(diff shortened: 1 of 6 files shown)


rhurey:
The change to Terser looks good.

The concern I have is changing the ECMA target is risk of breaking backward compatibility in a minor release instead of waiting for a 2.0.

rhurey:
If you want to pull the terser change into a separate PR, or scope this one down we can take it.

Domratchev:
@rhurey Both es5/commonjs and es2015/esnext modules are provided side-by-side.

Domratchev:
@rhurey I've rebased with your recent commit(s).
BTW, is there a reason you are not pulling ocsp library from npm?

rhurey:
I didn't find a way to get OCSP to work as-is if all network connections must be through a proxy server.

I could use https-proxy-agent or OCSP as the agent, but beyond that it started to turn into "duplicate a lot of OCSP's code" to get both to work together. My intent is to send a PR to OCSP that allows a user agent to be specified for cert checks and then revert to using the npm version if it's accepted.

Domratchev:
@rhurey Why don't you fork OCSP on GitHub and refer to it in dependencies like I do to this library?
E.g.
"dependencies": {
  "microsoft-cognitiveservices-speech-sdk": "github:Domratchev/cognitive-services-speech-sdk-js#mute"
}

Issue #39, opened 2019-03-12
 SDK needs to implement ECMAScript 2017


Joll59:
#### SDK should support modern JS. 
As it stands, complex actions with the SDK is cumbersome, because it utilizes callbacks.

It would be beneficial for the SDK to utilize promises or async/await, lending to a more streamlined coding experience.

rhurey:
Moving to a newer ECMA version is under discussion.

It's not clear which version that move would be to yet, or how much of a breaking change in the API surface will happen as an immediate result.

rhurey:
While this has been longer in coming that I'd hoped the retirement of home grown promises has begun.

Phase 1 will be to plumb native promises to the API surface.
Phase 2 will be to change the API to return Promise objects.

experiment/v2.0.promises is the working branch, I'll try to keep it in more working than not condition.

oscholz:
@Joll59 We have added this to our backlog for our next major version. When we complete this backlog item, we will update this issue with the version information that will contain the functionality.
I will close this issue for the time being.
Thank you for using the JS Speech SDK!

glharper:
@Joll59 We're currently thinking the SDK will wait for our v2.0 to implement ES2018 or higher. Upgrading now would be a breaking change for our customers on earlier versions of Node. Thanks again for using the JS Speech SDK, and we'll ping this issue when v2.0 gets closer (current timeline is TBD).

PR #58, opened 2019-04-17
 Migrate to es6 Promise. Add fetchToken support.

### LINKED ISSUE #57: Token based authentication is broken in Safari
There appears to be a feature (bug?) in the latest Safari (v12.1) where an attempt to resume AudioContext upon completion of an async operation fails, even it was initiated by a user action.
This makes it impossible to effectively use AuthorizationToken, as it only makes sense to fetch one upon a user action (e.g. button click) and the fetch has to be asynchronous. Plus there is no way to wait till AudioContext is resumed before supplying the token.

Proposed solution: CognitiveSubscriptionKeyAuthentication class already supports fetch (and fetchOnExpiry) function as a parameter. If it could be passed in via SpeechConfig, the issue would be solved. One obvious obstacle - fetch function returns proprietary Promise. Hence the part of the solution has to be the migration to es6 Promise.

Domratchev:
Resolves #57


rhurey:
First, thanks. Moving to ES6 Promises and getting the ability to write async / await code has been something on the "want" list for quite a while.

I'd started down migrating to native promises back in March when #39 was filed, and hadn't pushed the branch yet.

There were a few reasons that I held off over:
1) I wanted to be able to push the change all the way through the API surface and change to use Promise instead of the continuation pattern. But the time wasn't right for a change like that. (We want to keep the various SDK language bindings aligned in versioning and a breaking API surface change would require us to go to v2.0)

2) The risk of breaking something in the implementation seemed too high for a minor version release.

I'll take an in-depth look through the changes in the next couple of days, but my initial comments are that we can't move the property access methods around on the config objects.

Also, some behavior difference exists between the Promise implementations as there's a good sized chunk of tests failing now.

src/sdk/SpeechTranslationConfig.ts
```diff
@@ -92,26 +91,6 @@ export abstract class SpeechTranslationConfig extends SpeechConfig {
...
-    public abstract set authorizationToken(value: string);
-
-    /**
-     * Gets/Sets the speech recognition language.
-     * @member SpeechTranslationConfig.prototype.speechRecognitionLanguage
-     * @function
-     * @public
-     * @param {string} value - The authorization token.
-     */
-    public abstract set speechRecognitionLanguage(value: string);
```
rhurey: We can't remove methods from the SDK API surface without a major version revision, and we're not ready to move the SDK to 2.0 yet, so these should stay.

src/sdk/SpeechTranslationConfig.ts
```diff
@@ -147,24 +126,6 @@ export abstract class SpeechTranslationConfig extends SpeechConfig {
...
-     */
-    public abstract setProperty(name: string, value: string): void;
-
-    /**
-     * Dispose of associated resources.
-     * @member SpeechTranslationConfig.prototype.close
-     * @function
-     * @public
-     */
-    public abstract close(): void;
```
rhurey: Same comment as above about removing API members.

src/sdk/SpeechTranslationConfig.ts
Domratchev: SpeechTranslationConfig extends SpeechConfig which already declares/defines these properties and methods.

src/sdk/SpeechTranslationConfig.ts
Domratchev: SpeechTranslationConfig extends SpeechConfig which already declares/defines these properties and methods.
There is no change in the exposed API.

Domratchev:
@rhurey Ryan, do you see the migration to ES6 promises happening any time soon? I don't think that authorization token issue in Safari has any other solution.

orgads:
Can this be closed now?

glharper:
Closing due to inactivity.

PR #68, opened 2019-05-20
Remove throttling for the first 5 seconds of audio. 

rhurey:
After that throttle at 2x realtime.
Tracking the amount of data sent on the request session to each reconnection will fast flush any data.

Added more logging to tests to reduce the number of timeouts when things to wrong and get better error messages back.

### DIFF
--- a/src/common.speech/ServiceRecognizerBase.ts
+++ b/src/common.speech/ServiceRecognizerBase.ts
@@ -556,9 +556,13 @@ export abstract class ServiceRecognizerBase implements IDisposable {
         const deferred = new Deferred<boolean>();
 
         // The time we last sent data to the service.
-        let lastSendTime: number = Date.now();
+        let nextSendTime: number = Date.now();
 
         const audioFormat: AudioStreamFormatImpl = this.privAudioSource.format as AudioStreamFormatImpl;
+
+        // Max amount to send before we start to throttle
+        const fastLaneSizeMs: string = this.privRecognizerConfig.parameters.getProperty("SPEECH-TransmitLengthBeforThrottleMs", "5000");
+        const maxSendUnthrottledBytes: number = audioFormat.avgBytesPerSec / 1000 * parseInt(fastLaneSizeMs, 10);
 
         const readAndUploadCycle = () => {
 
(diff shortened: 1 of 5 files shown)


src/common.speech/ServiceRecognizerBase.ts
```diff
@@ -556,10 +556,14 @@ export abstract class ServiceRecognizerBase implements IDisposable {
         const deferred = new Deferred<boolean>();
 
         // The time we last sent data to the service.
-        let lastSendTime: number = Date.now();
+        let nextSendTime: number = Date.now();
 
         const audioFormat: AudioStreamFormatImpl = this.privAudioSource.format as AudioStreamFormatImpl;
 
+        // Max amount to send before we start to throttle
+        const fastLaneSizeMs: string = this.privRecognizerConfig.parameters.getProperty("SPEECH-TransmitLengthBeforThrottleMs", "5000");
```
fmegen: >y("SPEECH-TransmitLengthBeforThrottleMs", "5000"); [](start = 86, length = 50)

there should be an upper limit so a client cannot pass an arbitrary value here?

src/common.speech/ServiceRecognizerBase.ts
```diff
@@ -570,38 +574,53 @@ export abstract class ServiceRecognizerBase implements IDisposable {
...
+                            let sendDelay: number;
+
+                            if (audioStreamChunk.isEnd) {
+                                payload = null;
+                                sendDelay = 0;
+                            } else {
+                                payload = audioStreamChunk.buffer;
+                                this.privRequestSession.onAudioSent(payload.byteLength);
+
+                                if (maxSendUnthrottledBytes >= this.privRequestSession.bytesSent) {
```
fmegen: >maxSendUnthrottledBytes >= this.privRequestSession.bytesSent [](start = 36, length = 60)

in case the current sent operation crosses the unthrottled boundary, the transaction should be split into the part that gets send immediately and the other part that is throttled.
if not, the services may kick the client in case the part that is passed the unthrottled boundary is too large.

rhurey: This defaults the other way, so that if this frame crosses the boundary it's throttled.

---
In reply to: [286706670](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/68#discussion_r286706670) [](ancestors = 286706670)

PR #70, opened 2019-06-14
Add OCSP certificate validation when connecting from node.js

rhurey:
Also via proxy & Node.js

Imported the OCSP library from https://github.com/indutny/ocsp with a modification to allow the OCSP validation to work via proxy. (Will be working out of band to get that change contributed back and then revert to consuming directly instead.)

Enabled the SpeechReco tests to all run via proxy through config setting.

Changed the gulpfile to use the tsconfig.json in the root directory instead of having a separate copy of the settings.


package.json
```diff
@@ -79,7 +84,10 @@
     "usePathForSuiteName": "true"
   },
   "dependencies": {
+    "asn1.js-rfc2560": "^5.0.0",
+    "asn1.js-rfc5280": "^3.0.0",
```
fmegen: you need both rfcs?

rhurey: They're both used by the OCSP code I'm importing.

PR #75, opened 2019-07-25
OCSP: Disable when NODE_TLS_REJECT_UNAUTHORIZED is set to 0

orgads:
I'm not sure if this is the right way to solve it. Feel free to suggest alternatives :)

### DIFF
--- a/src/common.browser/WebsocketMessageAdapter.ts
+++ b/src/common.browser/WebsocketMessageAdapter.ts
@@ -98,6 +98,7 @@ export class WebsocketMessageAdapter {
         this.privConnectionState = ConnectionState.Connecting;
 
         try {
+            const enableOCSP: boolean = process.env.NODE_TLS_REJECT_UNAUTHORIZED !== "0";
             if (typeof WebSocket !== "undefined" && !WebsocketMessageAdapter.forceNpmWebSocket) {
                 // Browser handles cert checks.
                 this.privCertificateValidatedDeferral.resolve(true);
@@ -115,8 +116,10 @@ export class WebsocketMessageAdapter {
                     if (undefined !== this.proxyInfo.UserName) {
                         httpProxyOptions.headers = {
                             "Proxy-Authentication": "Basic " + new Buffer(this.proxyInfo.UserName + ":" + (this.proxyInfo.Password === undefined) ? "" : this.proxyInfo.Password).toString("base64"),
-                            "requestOCSP": "true",
                         };
+                        if (enableOCSP) {
+                            httpProxyOptions.headers.requestOCSP = "true";
+                        }
                     }
 
                     const httpProxyAgent: HttpsProxyAgent = new HttpsProxyAgent(httpProxyOptions);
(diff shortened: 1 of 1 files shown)


external/ocsp/ocsp/agent.js
```diff
@@ -67,6 +67,10 @@ Agent.prototype.createConnection = function createConnection(port,
 Agent.prototype.handleOCSPResponse = function handleOCSPResponse(socket,
                                                                  stapling,
                                                                  cb) {
+  if (process.env.NODE_TLS_REJECT_UNAUTHORIZED === '0') {
+    cb();
+    return;
```
rhurey: This should likely go in WebSocketMessageAdapter to not register the agent in the first place if checks are going to be skipped.

There's two reasons for this:
1) I'd like to keep as close to the ocsp library as possible in the hopes of getting the changes made so far committed back to the origin.

2) This doesn't cover the scenario where a proxy is used.

orgads: Ok, I pushed a revised commit. Is it better?

src/common.browser/WebsocketMessageAdapter.ts
```diff
@@ -95,6 +95,7 @@ export class WebsocketMessageAdapter {
         this.privConnectionState = ConnectionState.Connecting;
 
         try {
+            const enableOCSP = process.env.NODE_TLS_REJECT_UNAUTHORIZED !== "0";
```
rhurey: >const enableOCSP [](start = 12, length = 16)

const enableOCSP :boolean

src/common.browser/WebsocketMessageAdapter.ts
```diff
@@ -123,34 +126,37 @@ export class WebsocketMessageAdapter {
...
                 } else {
-                    // The ocsp library will handle validation for us and fail the connection if needed.
-                    this.privCertificateValidatedDeferral.resolve(true);
-
-                    const ocspAgent: ocsp.Agent = new ocsp.Agent({});
-                    const options: ws.ClientOptions = { agent: ocspAgent };
+                    const options: ws.ClientOptions = {};
+                    if (enableOCSP) {
+                        // The ocsp library will handle validation for us and fail the connection if needed.
+                        this.privCertificateValidatedDeferral.resolve(true);
```
rhurey: Ran the test suite with NODE_TLS_REJECT_UNAUTHORIZED = 0 and all the tests that use the service failed. 

I suspect it's due to this deferral never being resulted if enable OCSP is false. For the non-proxy case it should always be resolved, and in the proxy case resolved if enableOCSP is false.

rhurey:
It's been 4 months, I'm going to reject this PR if there isn't a response to comments in a week.

orgads:
Hi, sorry for the delay. I'll try to finish it this week.

orgads:
Next step is to make it configurable (possibly with default set according to `NODE_TLS_REJECT_UNAUTHORIZED`), but I'm not familiar with the code, so I'll leave it to you. :)

orgads:
@rhurey: My turn to ping you ;)

rhurey:
Closing and reopening to clear some Github->DevOps integration snags.

rhurey:
https://msasg.visualstudio.com/Skyman/_build?definitionId=7863 Passed on merge, not sure why GitHub integration won't see it. Still need to test with env enabled.

Issue #74, opened 2019-07-10
Too few chunks send to speech server in throtteled / minimized browser


LMIT-Dev:
Hi,

we've encountered an issue where an user using a speech to text app would not see any recognized data. This happens when the browser is minimized (in throtteled mode). As soon as the browser is pulled up again, the microphone data is send to the speech server.

We've tracked the issue down to the class [ServiceRecognizerBase](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/master/src/common.speech/ServiceRecognizerBase.ts) and its method "sendAudio". This method creates a function "readAndUploadCycle" that is then committed into a Timeout ("setTimeout"). The timout interval is calculated to provide a good send timeout. However, in a throtteled JS execution environment, this timeout interval is ignored and approximately only 1 chunk per second will be sent resulting in a huge and fast growing delay.

We believe this method should actually send multiple chunks to account for throtteled JS execution.

We hope this information is sufficient and would of course like to see an improvment of that logic soon.

For now we've asked our users to keep the browser window open in the background and not minimize it.

Anyways, we would also like to take the opportunity to thank you for the great work and offer our assistance if needed / desired. We are considering to create a fork to implement a logic suited for a throtteled environment if you feel that this issue isn't worth of improvements or are too busy for a near future investigation.

Thank you in advance for taking the time to consider this issue.
If possible please respond as soon as its possible.

LMIT-Dev:
To resolve this:

`
                                    setTimeout(function () {
                                        lastSendTime = Date.now();
                                        readAndUploadCycle();
                                    }, delay_1);
`
=>
`
                                    if (audioStreamNode.privBuffers.length == 1) {
                                        setTimeout(function () {
                                            lastSendTime = Date.now();
                                            readAndUploadCycle();
                                        }, delay_1);
                                    } else {
                                        new Promise(readAndUploadCycle);
                                    }
`

rhurey:
Well, that's not good. Thanks for narrowing down the problem. Thanks a lot.

I was thinking of solutions over the last day, and there's some complications in most of them.

[...]

PR #82, opened 2019-08-22
Fix Sending 1 chunk per second with Microphone in Throtteled JS environment

LMIT-Dev:
Fix Sending 1 chunk per second with Microphone in Throtteled JS
environment
Add comments to modified sections
Issue #74

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2836,8 +2836,7 @@
           "version": "2.1.1",
           "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
           "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -2861,15 +2860,13 @@
           "version": "1.0.0",
           "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
           "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
           "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
           "dev": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
@@ -2886,22 +2883,19 @@
           "version": "1.1.0",
           "resolved": "https://registry.npmjs.org/code-point-at/-/code-point-at-1.1.0.tgz",
           "integrity": "sha1-DQcLTQQ6W+ozovGkDi7bPZpMz3c=",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "concat-map": {
           "version": "0.0.1",
           "resolved": "https://registry.npmjs.org/concat-map/-/concat-map-0.0.1.tgz",
           "integrity": "sha1-2Klr13/Wjfd5OnMDajug1UBdR3s=",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "console-control-strings": {
           "version": "1.1.0",
           "resolved": "https://registry.npmjs.org/console-control-strings/-/console-control-strings-1.1.0.tgz",
           "integrity": "sha1-PXz0Rk22RG6mRL9LOVB/mFEAjo4=",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "core-util-is": {
           "version": "1.0.2",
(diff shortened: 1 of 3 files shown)


LMIT-Dev:
Apparently there is something failing:
image
But we cannot inspect it:
image
I am pretty sure we do not have access to the Azure portal. And I dont think we should have.

Regards

rhurey:
I was waiting until the CLA agreement was signed to do much digging. I've kicked off another test run and will share the results.

rhurey:
The tests around Silence are reporting a different number of NoMatchs coming back:
  ● Service based tests › Silence After Speech

    expect(received).toEqual(expected) // deep equality

    Expected: 0
    Received: 1

      1767 |                 r.stopContinuousRecognitionAsync(() => {
      1768 |                     try {
    > 1769 |                         expect(speechEnded).toEqual(noMatchCount);
           |                                             ^
      1770 |                         expect(noMatchCount).toEqual(2);
      1771 |                         done();
      1772 |                     } catch (error) {

      at tests/SpeechRecognizerTests.ts:1769:45
      at SpeechRecognizer.stopContinuousRecognitionAsync (src/sdk/SpeechRecognizer.ts:3281:11)
      at tests/SpeechRecognizerTests.ts:1767:19
      at WaitForCondition (tests/Utilities.ts:214:5)
      at tests/Utilities.ts:223:14
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:678:19)

  ● Service based tests › Silence Then Speech

    expect(received).toEqual(expected) // deep equality

    Expected: 3
    Received: 2

      1849 |                 r.stopContinuousRecognitionAsync(() => {
      1850 |                     try {
    > 1851 |                         expect(speechEnded).toEqual(noMatchCount + 1);
           |                                             ^
      1852 |                         expect(noMatchCount).toEqual(2);
      1853 |                         done();
      1854 |                     } catch (error) {

      at tests/SpeechRecognizerTests.ts:1851:45
      at SpeechRecognizer.stopContinuousRecognitionAsync (src/sdk/SpeechRecognizer.ts:3281:11)
      at tests/SpeechRecognizerTests.ts:1849:19
      at WaitForCondition (tests/Utilities.ts:214:5)
      at tests/Utilities.ts:223:14
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:678:19)

We're also sending more data than expected to get an initial silence response for translation:
  ● Service based tests › InitialSilenceTimeout (pull)

    expect(received).toBeLessThanOrEqual(expected)

    Expected: <= 646272
    Received:    819200

      960 |             // We should have sent 5 seconds of audio unthrottled and then 2x the time reco took until we got a response.
      961 |             const expectedBytesSent: number = (5 * 16000 * 2) + (2 * elapsed * 32000 / 1000);
    > 962 |             expect(bytesSent).toBeLessThanOrEqual(expectedBytesSent);
          |                               ^
      963 | 
      964 |         });
      965 |     }, 15000);

      at tests/TranslationRecognizerTests.ts:962:31
      at tests/TranslationRecognizerTests.ts:1047:21
      at WaitForCondition (tests/Utilities.ts:214:5)
      at tests/Utilities.ts:223:14
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:678:19)

LMIT-Dev:
are those tests done via microphone or file?

Might it be possible that the tests wont work since the nature of sending chunks changed for microphone input?

In our production environment, the microphone input is "translated" just fine, when we have time on our hands we can look into it again.

Could you share how to run these tests or link to a documentation (i didnt really look for anything but it would reduce the time we have to invest)

LMIT-Dev:
Closing this PR as noone seems to be interested in solving this...

rhurey:
I'm sorry that I didn't get back with instructions. Somehow I dropped the ball on this. Not much else besides excuses I can offer.

I am interested in getting this PR in.

None of our automated tests use a microphone, they're all based on audio files that are either in the repro or silence they manufacture internally with empty stream buffers. The microphone testing we do is a manual process.

The simplest way to run the tests is to use the runtests.cmd file in the root of the repro passing in the secrets required to get the tests to execute. The cmd line is like:

runtests.cmd SpeechSubscriptionKey:<KEY>SpeechRegion:<REGION> LuisSubscriptionKey:<LUISKEY> LuisRegion:westus BotSecret:<BOTKEY> BotSubscription:<BOTSUB> BotRegion:<BOTREGION>

Since the failures reproduce in the speech recognition tests, You really only need the SpeechSubscriptionKey and SpeechRegion parameters. You could either ignore the errors from the other suites, or scope the out by editing down the jest.config.js file's regex.

I dug into _Service based tests › Silence Then Speech_ to see why it was failing. It looks to be a timing bug around speech replay near the end of the stream.

The tests sends ~30 seconds of silence and then a wave file using a push stream. 

What should happen is every 15s of silence the service will end the turn and ignore all future audio with the same request ID, the SDK should see the end of turn, realize the service has stopped listening, change the request ID, back up the audio pointer in the replay buffer to the offset when the turn ended and continue to send audio. The service will then recognize the next turn. The test passes with the phrase in the wave file is recognized and the end of stream is reached.

What's happening is the phrase is never being recognized and the recognition is canceling after the 2 silence resets. I suspect some timing has changed in when the promise returned by sendAudio is resolved and the rewind that should be happening isn't because sendAudio thinks its already send the end of stream and its done, not realizing it needs to keep sending for at least one more loop.

Again, sorry I dropped the ball on this. We've been evolving our github / Stack overflow / etc processes to improve accountability on getting questions answered, this is an example of why.

PR #83, opened 2019-08-24
Fix error when quote exceeded

rhurey:
When a subscription exceeds its quota, continuous recognition doesn't stop and report the correct error.

Now, cancel the recognition and return the correct error code.

Also, found a bug where when start / stop / start was called in rapid succession the audio pump from the first start could continue to run based on timing.

### DIFF
--- a/src/common.speech/RequestSession.ts
+++ b/src/common.speech/RequestSession.ts
@@ -37,8 +37,8 @@ export class RequestSession {
     private privLastRecoOffset: number = 0;
     private privHypothesisReceived: boolean = false;
     private privBytesSent: number = 0;
-
-    protected privSessionId: string;
+    private privRecogNumber: number = 0;
+    private privSessionId: string;
 
     constructor(audioSourceId: string) {
         this.privAudioSourceId = audioSourceId;
@@ -75,6 +75,10 @@ export class RequestSession {
         return this.privTurnStartAudioOffset;
     }
 
+    public get recogNumber(): number {
+        return this.privRecogNumber;
+    }
+
     // The number of bytes sent for the current connection.
     // Counter is reset to 0 each time a connection is established.
     public get bytesSent(): number {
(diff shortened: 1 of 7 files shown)


lisaweixu:
>         if (!!this.privIsRecognizing) {

why !! ? not getting this logic.

---
Refers to: src/common.speech/RequestSession.ts:210 in 320afc2. [](commit_id = 320afc2a6b8d5499717c384e39773b7702ddf718, deletion_comment = False)

src/common.speech/RequestSession.ts
```diff
@@ -186,7 +191,7 @@ export class RequestSession {
     }
 
     public onStopRecognizing(): void {
-        this.privIsRecognizing = false;
+        this.onComplete();
```
lisaweixu: it seems the onComplete is called when turn ends. etc. Add this seems not hurting. Does it necessary?

src/common.speech/SpeechServiceRecognizer.ts
```diff
@@ -92,28 +92,12 @@ export class SpeechServiceRecognizer extends ServiceRecognizerBase {
                 if (ResultReason.Canceled === resultReason) {
                     const cancelReason: CancellationReason = EnumTranslation.implTranslateCancelResult(simple.RecognitionStatus);
 
-                    result = new SpeechRecognitionResult(
-                        this.privRequestSession.requestId,
-                        resultReason,
+                    this.cancelRecognitionLocal(
```
lisaweixu: >cancelRecognitionLocal [](start = 25, length = 22)

cancelRecognitinLocal is an interesting name. It is a kind of wrapper for the real impl in speech, translator and intent.

It seems you are calling stopRecognizing multiple times. Since JS is single threaded, i assume you are fine.

rhurey:
>         if (!!this.privIsRecognizing) {

It's likely less important in this instance as the underlying value *should* be a boolean, but in JS to do a boolean operator on a non-boolean type this will convert the value to a boolean.

---
In reply to: [531383426](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/83#issuecomment-531383426) [](ancestors = 531383426)

---
Refers to: src/common.speech/RequestSession.ts:210 in 320afc2. [](commit_id = 320afc2a6b8d5499717c384e39773b7702ddf718, deletion_comment = False)

src/common.speech/RequestSession.ts
rhurey: OnComplete also stops the audio pump when it detaches the audio node.

---
In reply to: [324360733](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/83#discussion_r324360733) [](ancestors = 324360733)

PR #88, opened 2019-09-14
Update properties & URL parameter support on config objects.

rhurey:
Add new properties to public SDK and stub methods to impl classes.

This brings the JS SDK into parity with the native implementation.

Minor test refactor to enable more tests.

Fixed a bug in the translation & intent recognizers where if the canceled event was not suscribed to,
recognizeOnceAsync would not indicate recognation had canceled due to an internal error.

### DIFF
--- a/src/common.speech/SpeechConnectionFactory.ts
+++ b/src/common.speech/SpeechConnectionFactory.ts
@@ -5,13 +5,32 @@ import {
     ProxyInfo,
     WebsocketConnection,
 } from "../common.browser/Exports";
-import { OutputFormatPropertyName } from "../common.speech/Exports";
-import { IConnection, IStringDictionary } from "../common/Exports";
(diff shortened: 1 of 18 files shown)


src/common.speech/SpeechConnectionFactory.ts
```diff
@@ -5,13 +5,32 @@ import {
...
+    RecognitionMode,
+    RecognizerConfig,
+    WebsocketMessageFormatter
+} from "./Exports";
+import {
+    QueryParameterNames
+} from "./QueryParameterNames";
 
-export class SpeechConnectionFactory implements IConnectionFactory {
+export class SpeechConnectionFactory extends ConnectionFactoryBase {
```
lisaweixu: do you want to add the newly added SendMessageSync and SetMessageProperty in the native class of connection to here as well?

PR #89, opened 2019-09-19
Don't reuse variable names in telemetry verfication

rhurey:


tests/TelemetryUtil.ts
```diff
@@ -4,16 +4,18 @@
...
             phrases = telemetryMessage.ReceivedMessages["translation.phrase"];
         }
-        expect(phrases.length).toEqual(phrases);
+        expect(phrases).not.toBeUndefined();
+        expect(phrases.length).toEqual(numPhrases);
     }
+
     if (0 > hypothesis) {
         const hypothesis: string[] = telemetryMessage.ReceivedMessages["speech.hypothesis"];
         expect(hypothesis.length).toEqual(hypothesis);
```
BrianMouncer: >expect(hypothesis.length).toEqual(hypothesis); [](start = 8, length = 46)

same issue here?

you have a count and an array with the same name for "hypothesis" and do the same size to array compare instead of count to length....

rhurey: Yup. Good catch. Thanks.

---
In reply to: [326432971](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/89#discussion_r326432971) [](ancestors = 326432971)

PR #97, opened 2019-10-23
Fix machine in the middle vulnerability for "https-proxy-agent"

gghizila:
- changes generated by npm audit fix --force
- upgrade "https-proxy-agent" to 3.0.0 as there is inconsistency on reports as what version fixes the vulnerability

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1366,7 +1366,7 @@
         "component-emitter": "^1.2.1",
         "define-property": "^1.0.0",
         "isobject": "^3.0.1",
-        "mixin-deep": "^1.3.2",
+        "mixin-deep": "^1.2.0",
         "pascalcase": "^0.1.1"
       },
       "dependencies": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -1701,7 +1701,7 @@
         "get-value": "^2.0.6",
         "has-value": "^1.0.0",
         "isobject": "^3.0.1",
-        "set-value": "^2.0.1",
+        "set-value": "^2.0.0",
```
BrianMouncer: >"set-value": "^2.0.0", [](start = 8, length = 22)

I am surprised we would ever go back a version, and a previous version of "set-value" have vulnerability we recently had to move up to mitigate, so please make sure this was needed to revert to a previous version.

gghizila: There is something odd with npm, audit command or otherwise. It picked these values based on the new dependency. Why it downgraded existing ones, brought in I suppose by just pulling a top level dependency, beats me.

Issue #103, opened 2019-11-14
PropertyId.Conversation_From_Id does not seem to be supported


alopix:
There does not seem to be support for `PropertyId.Conversation_From_Id` yet – at least setting it does not do anything and it's even necessary to cast to `any` in order to set  the property.
```
this.dialogServiceConfig.setProperty(
  PropertyId.Conversation_From_Id as any,
  'my value'
);
```

Am I using it wrong or was it not implemented yet?

modnarb:
@alopix not using it wrong, I've just published a PR ([110](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/110)) that contains the fix to address this.

modnarb:
I've merged the change, closing issue.

PR #110, opened 2019-12-06
Fix for audio turn processing and add support for missing properties

modnarb:
- Fix for audio turns being corrupted from reco result returning after turn end
- Add fromId support
- Add outputformat support, add to query params
- Protect against OOM when starting turns before the previous one completed

### DIFF
--- a/src/common.speech/AgentConfig.ts
+++ b/src/common.speech/AgentConfig.ts
@@ -28,7 +28,8 @@ export interface IAgentConfig {
     botInfo: {
         commType: string,
         connectionId: string,
-        conversationId: string
+        conversationId: string,
+        fromId: string
     };
     version: number;
 }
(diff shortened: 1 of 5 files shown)


gghizila:
Can you change the description of this PR?

src/common.speech/AgentConfig.ts
```diff
@@ -28,7 +28,8 @@ export interface IAgentConfig {
     botInfo: {
         commType: string,
         connectionId: string,
-        conversationId: string
+        conversationId: string,
+        fromId: string
```
gghizila: I am not sure where is this used.

modnarb: It's a property that can be set by the customer. WebChat requested it as did the user in this post https://github.com/microsoft/cognitive-services-speech-sdk-js/issues/103

modnarb:
> Can you change the description of this PR?

Done

modnarb:
Re-ran build and all tests pass, #20191206.3

PR #113, opened 2019-12-17
Honor service property ids when set

modnarb:
Adding a call to the factory base method that parses out specified service properties, as added in 1.7

While the usage of these may change soon, for now this is the only way to specify "service properties" so we need this change to support it.

usage example:
config.setServiceProperty("some_propertyName", "some_value", SpeechSDK.ServicePropertyChannel.UriQueryParameter)


modnarb:
Note that I've intentionally not added the service property Ids to the enum dictionary.

PR #115, opened 2020-01-10
Send a wave header based on the AudioFormat specified.

rhurey:
The change to MTS service side has increased enforcement of the data in the WAV header, and the SDK needs to send a correct header.

This will send one based on the audio stream format specified when a stream is created.

### DIFF
--- a/tests/WaveFileAudioInputStream.ts
+++ b/tests/WaveFileAudioInputStream.ts
@@ -19,7 +19,7 @@ export class WaveFileAudioInput {
     public static LoadArrayFromFile(filename: string): ArrayBuffer {
         const fileContents: Buffer = fs.readFileSync(filename);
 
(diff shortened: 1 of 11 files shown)


tests/WaveFileAudioInputStream.ts
```diff
@@ -19,7 +19,7 @@ export class WaveFileAudioInput {
     public static LoadArrayFromFile(filename: string): ArrayBuffer {
         const fileContents: Buffer = fs.readFileSync(filename);
 
-        const ret = Uint8Array.from(fileContents);
+        const ret = Uint8Array.from(fileContents.slice(44));
```
lisaweixu: Some wav headers are not 44 bytes long e.g. more than 2 channels pcm file normally has more than 44 bytes. Please add a comment saying for now only support 1 channel pcm data as wav.

PR #117, opened 2020-01-13
Fix reconnect spamming

modnarb:
In the current design, the message listener and connection configuration can get into an error state where they repeatedly call one another, resulting in continuous attempts to connect. Ultimately this causes an out-of-memory on the client.

Now, if the termination of the connection/messaging loop has been flagged it will take effect before the next attempt to reconnect.

### DIFF
--- a/src/common.speech/DialogServiceAdapter.ts
+++ b/src/common.speech/DialogServiceAdapter.ts
@@ -508,6 +508,8 @@ export class DialogServiceAdapter extends ServiceRecognizerBase {
                     || this.privDialogConnectionPromise.result().result.state() === ConnectionState.Disconnected)) {
                 this.agentConfigSent = false;
                 this.privDialogConnectionPromise = null;
+                this.terminateMessageLoop = true;
+                return this.configConnection();
             } else {
                 return this.privDialogConnectionPromise;
             }
@@ -717,6 +719,10 @@ export class DialogServiceAdapter extends ServiceRecognizerBase {
             } else {
                 return this.privConnectionConfigPromise;
             }
+        }
+
+        if (this.terminateMessageLoop) {
+            return PromiseHelper.fromResult<IConnection>(undefined);
         }
 
         this.privConnectionConfigPromise = this.dialogConnectImpl().onSuccessContinueWithPromise((connection: IConnection): Promise<IConnection> => {


src/common.speech/DialogServiceAdapter.ts
```diff
@@ -508,6 +508,8 @@ export class DialogServiceAdapter extends ServiceRecognizerBase {
                     || this.privDialogConnectionPromise.result().result.state() === ConnectionState.Disconnected)) {
                 this.agentConfigSent = false;
                 this.privDialogConnectionPromise = null;
+                this.terminateMessageLoop = true;
```
gghizila: Is this happening in the other connection state management for SR or intent etc? you may want to open a bug if that is the case. Interesting how are that they are not affected.

src/common.speech/DialogServiceAdapter.ts
```diff
@@ -719,6 +721,10 @@ export class DialogServiceAdapter extends ServiceRecognizerBase {
             }
         }
 
+        if (this.terminateMessageLoop) {
```
gghizila: this is set from two other places not just the newly added code, is this handled correctly for those cases as well?

gghizila:
Did you check why the tests failed as well?

src/common.speech/DialogServiceAdapter.ts
modnarb: The issue is present in SR, but SR is not affected to the same degree because they don't rely on having a persistently open connection, where messages can be sent/received at any time. For SR, they "gate" this message loop using the recognizer session state - if the recognizer is not in the "recognizing" state, they break the loop. But until that check is made the reconnect looping is occurring... It just never gets out of control since that check is made on the next loop.

src/common.speech/DialogServiceAdapter.ts
modnarb: Yes, I added those in an early mitigation attempt for this same issue manifesting in another way.

modnarb:
> Did you check why the tests failed as well?

Yes and I re-ran with a successful pass. I don't know how to schedule the re-run from the github repo or how to reflect that re-run to the merge criteria box below, but [here's the link](https://msasg.visualstudio.com/Skyman/_build/results?buildId=9078886)

rhurey:
You can schedule against refs/pull/#/head and github will figure it out. 

For some reason refs/pull/#/merge only works if github schedules it.

modnarb:
> You can schedule against refs/pull/#/head and github will figure it out.
> 
> For some reason refs/pull/#/merge only works if github schedules it.

Awesome, will try this out on my next run. thanks!

PR #127, opened 2020-01-27
Fix missing null check for Android WebView

### LINKED ISSUE #128: SpeechRecognizer throws undefined TypeError on Android WebView
I'm building an Ionic app and have been testing out using the JS SDK instead of building a custom native plugin.

I encountered an error when the audio is being shutdown:
Screen Shot 2020-01-26 at 9 26 08 PM

I was running a simple Ionic app that calls `recognizeOnceAsync` on the `SpeechRecognizer` using Android API 28 and Android API 29.

rozele:
Encountered an undefined reference while trying to get this working in an Ionic app on Android API 29 emulator. This fixes the undefined reference (and matches the approach taken in the [`DialogServiceAdapter`](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/d30a67c1f9319c72a3eba4446abfee12aa8d9d61/src/common.speech/DialogServiceAdapter.ts#L440)).

Fixes #128

### DIFF
--- a/src/common.speech/ServiceRecognizerBase.ts
+++ b/src/common.speech/ServiceRecognizerBase.ts
@@ -611,7 +611,7 @@ export abstract class ServiceRecognizerBase implements IDisposable {
                             let payload: ArrayBuffer;
                             let sendDelay: number;
 
-                            if (audioStreamChunk.isEnd) {
+                            if (!audioStreamChunk || audioStreamChunk.isEnd) {
                                 payload = null;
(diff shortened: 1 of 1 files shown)


rozele:
Here's a screenshot of the bug:
Screen Shot 2020-01-26 at 9 26 08 PM

rozele:
Any chance we can get this patched into a v1.8.2 or something? Having trouble using v1.9.x.

PR #134, opened 2020-01-30
Add content type to audio packets

rhurey:


src/common.speech/ServiceRecognizerBase.ts
```diff
@@ -475,7 +475,7 @@ export abstract class ServiceRecognizerBase implements IDisposable {
             MessageType.Binary,
             "audio",
             this.privRequestSession.requestId,
-            null,
+            "audio/x-wav",
```
gghizila: Will these changes send the content type on every message?

rhurey: Yes.

rhurey: Just rand Fiddler on the native client and it's only putting the content type on the header packet. I'll do the same.

PR #140, opened 2020-02-07
Read custom endpoint from config properties

modnarb:
Allows users to set a custom endpoint, else will default to the Dialog or CustomCommands endpoints

### DIFF
--- a/src/common.speech/DialogConnectorFactory.ts
+++ b/src/common.speech/DialogConnectorFactory.ts
@@ -71,18 +71,17 @@ export class DialogConnectionFactory extends ConnectionFactoryBase {
         }
         headers[QueryParameterNames.ConnectionIdHeader] = connectionId;
 
+        if (applicationId !== "") {
+            headers[authHeader] = applicationId;
+        }
+
         let endpoint: string = config.parameters.getProperty(PropertyId.SpeechServiceConnection_Endpoint, "");
         if (endpoint === "") {
-            // ApplicationId is only required for CustomCommands
+            // ApplicationId is only required for CustomCommands, so we're using that to determine default endpoint
             if (applicationId === "") {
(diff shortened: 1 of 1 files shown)


src/common.speech/DialogConnectorFactory.ts
```diff
@@ -66,18 +66,25 @@ export class DialogConnectionFactory extends ConnectionFactoryBase {
         const {resourcePath, version, authHeader} = getDialogSpecificValues(dialogType);
 
         const headers: IStringDictionary<string> = {};
-        headers[authInfo.headerName] = authInfo.token;
+        if (authInfo.token !== undefined && authInfo.token !== "") {
```
gghizila: don't you have to check for null as well since it comes from the API developer?

src/common.speech/DialogConnectorFactory.ts
```diff
@@ -66,18 +66,25 @@ export class DialogConnectionFactory extends ConnectionFactoryBase {
...
+        }
         headers[QueryParameterNames.ConnectionIdHeader] = connectionId;
 
-        let endpoint: string;
-        // ApplicationId is only required for CustomCommands
-        if (applicationId === "") {
-            endpoint = `wss://${region}.${baseUrl}/${pathSuffix}/${version}`;
-        } else {
-            endpoint = `wss://${region}.${baseUrl}/${resourcePath}/${pathSuffix}/${version}`;
+        if (applicationId !== "") {
```
gghizila: same comment as above with null and undefined

modnarb: This one will have a value or be empty, per the call above where it's set

src/common.speech/DialogConnectorFactory.ts
modnarb: Good call, replacing with a "juggling-check"

PR #146, opened 2020-02-13
Provide setting custom "activity template"

modnarb:


src/common.speech/DialogServiceAdapter.ts
```diff
@@ -792,6 +792,7 @@ export class DialogServiceAdapter extends ServiceRecognizerBase {
             context: {
                 interactionId: guid
             },
+            messagePayload: this.activityTemplate ? this.activityTemplate : undefined,
```
gghizila: What is the "persistence" story of the template? It is used for as long as is set and it needs to be reset?

src/sdk/DialogServiceConnector.ts
```diff
@@ -144,6 +144,10 @@ export class DialogServiceConnector extends Recognizer {
         return this.privProperties;
     }
 
+    public set activityTemplate(messagePayload: string) {
```
gghizila: So this is a property on the connector not on the config?

src/common.speech/DialogServiceAdapter.ts
gghizila: This sounds like your property is a string hence it will be a Json string and not a Json object. This is what I expect but I want to make sure this is what you wanted.

src/common.speech/ServiceRecognizerBase.ts
```diff
@@ -278,6 +279,9 @@ export abstract class ServiceRecognizerBase implements IDisposable {
 
     public sendMessage(message: string): void { }
 
+    public set activityTemplate(messagePayload: string) { this.privActivityTemplate = messagePayload; }
+    public get activityTemplate(): string { return this.privActivityTemplate; }
```
rhurey: Is this a base class thing?

modnarb: I set as such since we want access from the api surface as well as the recognizer, but as I "say" that out loud perhaps this is better served via the property collection?

src/sdk/DialogServiceConnector.ts
modnarb: Correct, this enables the client to "stamp" their activities and have that persist on following turns, until they set/reset it. Though on the topic of properties, maybe this better belongs in our property collection (per conversation w/Ryan, below)

src/common.speech/DialogServiceAdapter.ts
modnarb: Correct, this enables the client to "stamp" their activities and have that persist on following turns, until they set/reset it. Though on the topic of properties, maybe this better belongs in our property collection (per conversation w/Ryan, below)

PR #147, opened 2020-02-14
Add interfaces and initial tests for Multi-device conversation

jk-ms:
Start adding in support for the Multi-device Conversation feature:


src/sdk/Exports.ts
```diff
@@ -50,3 +50,17 @@ export { ActivityReceivedEventArgs } from "./ActivityReceivedEventArgs";
 export { ServicePropertyChannel } from "./ServicePropertyChannel";
 export { ProfanityOption } from "./ProfanityOption";
 export { BaseAudioPlayer } from "./Audio/BaseAudioPlayer";
+export { Conversation,
+    ConversationExpirationEventArgs,
+    ConversationImpl,
```
rhurey: >ConversationImpl [](start = 4, length = 16)

Why are we exporting a Impl class on the public SDK interface?

tests/ConversationTranslatorTests.ts
```diff
@@ -0,0 +1,105 @@
...
+    ConsoleLoggingListener
+} from "../src/common.browser/Exports";
+import {
+    Events,
+    EventType,
+} from "../src/common/Exports";
+import { Settings } from "./Settings";
+
+// tslint:disable-next-line:no-console
+const consoleInfo = console.info;
```
rhurey: Nice. I need to remember this....

src/sdk/Exports.ts
jk-ms: Updated code to remove this export as it's not required here.

src/sdk/Exports.ts
```diff
@@ -50,3 +50,17 @@ export { ActivityReceivedEventArgs } from "./ActivityReceivedEventArgs";
...
+    ConversationParticipantsChangedEventArgs,
+    ConversationTranslationCanceledEventArgs,
+    ConversationTranslationEventArgs,
+    ConversationTranslationResult,
+    ConversationTranslator,
+    IConversation,
```
rhurey: >IConversation [](start = 4, length = 13)

Why expose both the interface and the abstract class?

jk-ms: Do you mean for the Conversation / IConversation?
If I've got a Typescript client consuming the sdk, then I get a no exported member error if they aren't both exported, e.g.
const myConversation: sdk.IConversation = sdk.Conversation.createConversationAsync(myConfig);

jk-ms: The interface is removed from the exports.ts

src/sdk/ConversationTranslator/Conversation.ts
```diff
@@ -0,0 +1,161 @@
...
+import { PropertyCollection, SpeechTranslationConfig } from "../Exports";
+import { IConversation } from "./IConversation";
+import { IParticipant, IUser } from "./IParticipant";
+
+export abstract class Conversation implements IConversation {
+
+    public abstract get authorizationToken(): string;
+    public abstract set authorizationToken(value: string);
+
+    public abstract get config(): SpeechTranslationConfig;
```
ralph-msft: Please remove these. Setting the SpeechConfig after a conversation has started is not a supported operation. Getting the SpeechConfig and then changing a property there also leads to an expectation that would change the Conversation which it does not.

src/sdk/ConversationTranslator/Conversation.ts
```diff
@@ -0,0 +1,161 @@
...
+    public abstract set config(value: SpeechTranslationConfig);
```
ralph-msft: Please remove (see above)

PR #148, opened 2020-02-15
Stop Recognizing should wait for pending service messages

rhurey:
Instead of Stop meaning "Stop everything" change stop to mean "Stop sending audio, and allow the service to respond to send but unrecognized audio."


src/sdk/TranslationRecognizer.ts
```diff
@@ -267,34 +206,7 @@ export class TranslationRecognizer extends Recognizer {
...
-                } else {
-                    err(error);
-                }
-            }
-
-            // Destroy the recognizer.
-            this.dispose(true);
-
-        }
+        this.stopContinuousRecognitionAsyncImpl(cb, err);
```
lisaweixu: nice cleanup

PR #150, opened 2020-02-24
Make speech activity template a property bag item 

modnarb:
Moving the previously added "activity template" property to a defined item in PropertyId list.


src/sdk/DialogServiceConnector.ts
```diff
@@ -144,8 +144,21 @@ export class DialogServiceConnector extends Recognizer {
         return this.privProperties;
     }
 
-    public set activityTemplate(messagePayload: string) {
```
rhurey: > activityTemplate [](start = 14, length = 17)

Is making this a breaking change (The name changed) intentional?

modnarb: "Yes".. meaning it's not really a breaking change, this was only added last week so shouldn't be out in the wild except for anyone who built a private drop.

PR #161, opened 2020-03-20
TTS implementation

yulin-li:
initial commit for TTS implementation, with NodeJS support only.


src/sdk/SpeechSynthesizer.ts
```diff
@@ -0,0 +1,419 @@
...
+    SpeechSynthesisEventArgs,
+    SpeechSynthesisOutputFormat,
+    SpeechSynthesisResult,
+    SpeechSynthesisWordBoundaryEventArgs,
+} from "./Exports";
+import {SpeechConfig, SpeechConfigImpl} from "./SpeechConfig";
+
+/**
+ * Defines the class SpeechSynthesizer for text to speech.
+ * @class SpeechSynthesizer
```
yinhew: You need comment this:
Added in version 1.11.0

tests/DialogServiceConnectorTests.ts
```diff
@@ -362,24 +362,22 @@ describe.each([true, false])("Service-based tests", (forceNodeWebSocket: boolean
         //     }
         // };
 
+        const audioBuffer = new ArrayBuffer(320);
         const audioReadLoop = (audioStream: PullAudioOutputStream, done: jest.DoneCallback) => {
-            audioStream.read().on((audioBuffer: ArrayBuffer) => {
+            audioStream.read(audioBuffer).on((bytesRead: number) => {
```
modnarb: noted, thanks for updating

src/common.browser/AudioFileWriter.ts
```diff
@@ -0,0 +1,55 @@
...
+        if (this.privAudioFormat.hasHeader) {
+            this.privOffset = this.privAudioFormat.header.byteLength;
+        }
+    }
+
+    public write(buffer: ArrayBuffer): void {
+        Contracts.throwIfNullOrUndefined(this.privAudioFormat, "must set format before writing.");
+        if (this.privFd !== undefined) {
+            fs.writeSync(this.privFd, new Int8Array(buffer), 0, buffer.byteLength, this.privOffset);
+            this.privOffset += buffer.byteLength;
```
modnarb: this should be the length used, but writeSync also returns the bytes read - wondering if that's a safer convention to use?

src/common.speech/SpeechSynthesisConnectionFactory.ts
```diff
@@ -0,0 +1,55 @@
...
+    private readonly synthesisUri: string = "/cognitiveservices/websocket/v1";
+
+    public create = (
+        config: SynthesizerConfig,
+        authInfo: AuthInfo,
+        connectionId?: string): IConnection => {
+
+        let endpoint: string = config.parameters.getProperty(PropertyId.SpeechServiceConnection_Endpoint, undefined);
+        const region: string = config.parameters.getProperty(PropertyId.SpeechServiceConnection_Region, undefined);
+        const hostSuffix = region.startsWith("china") ? ".azure.cn" : ".microsoft.com";
```
rhurey: >hostSuffix = region.startsWith("china") ? ".azure.cn" : ".microsoft.com"; [](start = 14, length = 73)

I think the rest of the factories also need this change.

src/common.speech/SynthesisAdapterBase.ts
```diff
@@ -0,0 +1,630 @@
...
+    // A promise for a configured connection.
+    // Do not consume directly, call fetchConnection instead.
+    private privConnectionConfigurationPromise: Promise<IConnection>;
+
+    // A promise for a connection, but one that has not had the speech context sent yet.
+    // Do not consume directly, call fetchConnection instead.
+    private privConnectionPromise: Promise<IConnection>;
+    private privAuthFetchEventId: string;
+    private privIsDisposed: boolean;
+    private privMustReportEndOfStream: boolean;
```
rhurey: >privMustReportEndOfStream [](start = 12, length = 25)

Is this ever read?

src/common.speech/SynthesisAdapterBase.ts
```diff
@@ -0,0 +1,630 @@
...
+            return this.speakOverride(ssml, requestId, successCallback, errorCallBack);
+        }
+
+        this.privSuccessCallback = successCallback;
+        this.privErrorCallback = errorCallBack;
+
+        this.privSynthesisTurn.startNewSynthesis(requestId, text, isSSML, audioDestination);
+
+        // Start the connection to the service. The promise this will create is stored and will be used by configureConnection().
+        this.connectImpl();
```
rhurey: You can just call fetchConnection since you don't need to do something expensive like getting a microphone in parallel.

src/common.speech/SynthesisAdapterBase.ts
```diff
@@ -0,0 +1,630 @@
...
+
+        return this.fetchConnection().continueWithPromise<boolean>((connection: PromiseResult<IConnection>) => {
+            if (connection.isError) {
+                this.cancelSynthesisLocal(CancellationReason.Error, CancellationErrorCode.ConnectionFailure, connection.error);
```
rhurey: You need to return here also, right? Else it'll try to send the context on the error'd connection

src/sdk/Audio/AudioOutputStream.ts
```diff
@@ -61,22 +69,21 @@ export abstract class PullAudioOutputStream extends AudioOutputStream {
...
      * Reads audio data from the internal buffer.
      * @member PullAudioOutputStream.prototype.read
      * @function
      * @public
-     * @returns {Promise<ArrayBuffer>} Audio buffer data.
+     * @param {ArrayBuffer} dataBuffer - An ArrayBuffer to store the read data.
+     * @returns {Promise<number>} Audio buffer length has been read.
      */
-    public abstract read(): Promise<ArrayBuffer>;
+    public abstract read(dataBuffer: ArrayBuffer): Promise<number>;
```
rhurey: Is this a breaking change for existing users?

src/common.speech/SynthesisAdapterBase.ts
yulin-li: Yes, for sure it'll send the context. Thanks for pointing this, will fix.

src/sdk/Audio/AudioOutputStream.ts
yulin-li: Seems it's a breaking change, but I think this change is necessary. Can we have some workaround?

src/common.speech/SynthesisAdapterBase.ts
yulin-li: thx, I see. Will update.

src/common.speech/SynthesisAdapterBase.ts
yulin-li: ah, I copyed this from `ServiceRecognizerBase` and forget to clean it. Thanks

src/common.speech/SpeechSynthesisConnectionFactory.ts
yulin-li: I will update this change to other factories.

src/common.browser/AudioFileWriter.ts
yulin-li: Thanks for pointing this, it's not safe indeed. I will try to use `fs.WriteStream` instead.

tests/DialogServiceConnectorTests.ts
yulin-li: sorry for this breaking change

src/common.speech/SpeechSynthesisConnectionFactory.ts
yulin-li: will do it in another PR, resolving conversation here.

yulin-li: see #163

Issue #153, opened 2020-03-04
Can't use Custom Speech endpoint with DialogServiceConnector 


henryjenkins:
Using microsoft-cognitiveservices-speech-sdk@1.10.0.

We are trying to use a custom speech endpoint with DialogServiceConnector so that we can leverage a custom speech model with the Direct Line Speech Bot Framework channel.

We are setting the custom endpoint as a service property on the BotFrameworkConfig as follows:

```
botFrameworkConfig = BotFrameworkConfig.fromAuthorizationToken(
      authorizationToken,
      region
    );

botFrameworkConfig.setServiceProperty(
        'cid',
        ourCustomEndpointId,
        ServicePropertyChannel.UriQueryParameter
      );
```

With this property set, speech recognition stops working. 
We receive the session started event, but not the speech start detected or recognizing/recognized events.
Removing the property restores normal operation.

The same code and endpoint appears to work correctly with a C# client.

rhurey:
@modnarb, can you take a look at this?

henryjenkins:
This seems to have been fixed in microsoft-cognitiveservices-speech-sdk@1.11.0.
Thanks @modnarb !

PR #165, opened 2020-03-27
Update casing on request headers

modnarb:
The custom speech service makes a case-sensitive check against header parameters, causing errors on any requests with mismatched casing.

The service should resolve this to be inline with the http standard https://tools.ietf.org/html/rfc2616#section-4.2

As a mitigation step, I've changed the request headers to use the casing that the speech service is expecting per https://speechwiki.azurewebsites.net/partners/speechsdk.html#required-message-headers

This addresses https://github.com/microsoft/cognitive-services-speech-sdk-js/issues/153


src/common.speech/DialogServiceAdapter.ts
```diff
@@ -501,7 +501,7 @@ export class DialogServiceAdapter extends ServiceRecognizerBase {
                 MessageType.Binary,
                 "audio",
                 this.privDialogRequestSession.requestId,
-                null,
+                "audio/x-wav",
```
modnarb: Wondering about removing this change.. could this cause any issues for future formats? I've just tested and it works with 'null'

rhurey: It depends on the endpoint and how strict it is when it transforms the audio.

The native code sends the header.

The JS code got burned when some endpoints' transformation service started demanding the header.

If we stop transmitting PCM, we're going to have a lot of work to do anyways around the compression.

---
In reply to: [399573404](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/165#discussion_r399573404) [](ancestors = 399573404)

PR #172, opened 2020-04-09
[tts] support browser

yulin-li:


src/sdk/Exports.ts
```diff
@@ -69,3 +69,4 @@ export { SpeechSynthesizer } from "./SpeechSynthesizer";
 export { SpeechSynthesisResult } from "./SpeechSynthesisResult";
 export { SpeechSynthesisEventArgs } from "./SpeechSynthesisEventArgs";
 export { SpeechSynthesisWordBoundaryEventArgs} from "./SpeechSynthesisWordBoundaryEventArgs";
+export { SpeakerAudioDestination } from "../common.browser/SpeakerAudioDestination";
```
rhurey, line 72: Objects that the SDK is going to expose should be in the SDK folder.

All the other folders are internal details.

src/sdk/Audio/AudioConfig.ts
```diff
@@ -114,6 +114,27 @@ export abstract class AudioConfig {
...
+     * Creates an AudioConfig object representing the custom IPlayer object.
+     * You can use the IPlayer object to control pause, resume, etc.
+     * @member AudioConfig.fromSpeakerOutput
+     * @function
+     * @public
+     * @param {IPlayer} player - the IPlayer object for playback.
+     * @returns {AudioConfig} The audio output configuration being created.
+     * Added in version 1.12.0
+     */
+    public static fromSpeakerOutput(player?: IPlayer): AudioConfig {
```
rhurey, line 127: >IPlayer [](start = 45, length = 7)

Do we need to export this if it's a type parameter.

src/sdk/Exports.ts
yulin-li: Thanks, move the `SpeakerAudioDestination` to `sdk/Audio`

src/sdk/Audio/AudioConfig.ts
yulin-li: I added some doc and export it.

PR #179, opened 2020-05-13
Fix firefox microphone input sample rate issue

glharper:
Account for FF behavior not allowing different sample rates when connecting AudioNodes


rhurey:
Ah, this should target master and then get cherry-picked into the release branch.

PR #183, opened 2020-05-22
[tts] support ogg format

yulin-li:
- support ogg format
- fix playback not working bug in Safari
- rename Jessa to Aria


rhurey:
>     public static buildSsml(text: string, properties: PropertyCollection): string {

Not for this PR, but long term this should likely come from a REST call.

---
Refers to: src/sdk/SpeechSynthesizer.ts:153 in b4e455b. [](commit_id = b4e455b5d3525a076720d99f9bf552be354c5cfe, deletion_comment = False)

PR #191, opened 2020-06-04
[tts] support language auto detection

yulin-li:
**do not merge now** waiting for service update


src/common.speech/SynthesisContext.ts
```diff
@@ -50,6 +51,9 @@ export class SynthesisContext {
                     wordBoundaryEnabled: (!!this.privSpeechSynthesizer.wordBoundary),
                 },
                 outputFormat: this.privAudioOutputFormat.requestAudioFormatString,
+            },
+            language: {
+                autoDetection: this.privSpeechSynthesizer.properties.getProperty(PropertyId.SpeechServiceConnection_AutoDetectSourceLanguages) === AutoDetectSourceLanguagesOpenRangeOptionName
```
glharper, line 56: Please make autoDetection a boolean get method from SpeechSynthesizer to avoid SynthesisContext having knowledge of propertyIds. This line should look like:

autoDetection: this.privSpeechSynthesizer.autoDetection

(This also means SynthesisContext doesn't need to import AutoDetectSourceLanguagesOpenRangeOptionName)

src/sdk/AutoDetectSourceLanguageConfig.ts
```diff
@@ -11,6 +12,7 @@ import {
 /**
  * Language auto detect configuration.
  * @class AutoDetectSourceLanguageConfig
+ * Added in version 1.13.0.
```
glharper, line 15: good catch

tests/SpeechSynthesisTests.ts
```diff
@@ -612,4 +623,53 @@ describe.each([true, false])("Service based tests", (forceNodeWebSocket: boolean
...
+        s.SynthesisCanceled = (o: sdk.SpeechSynthesizer, e: sdk.SpeechSynthesisEventArgs): void => {
+            done.fail();
+        };
+
+        // we will get very short audio as the en-US voices are not mix-lingual.
+        s.speakTextAsync("你好世界。", (result: sdk.SpeechSynthesisResult): void => {
+            // tslint:disable-next-line:no-console
+            console.info("speaking finished, turn 1");
+            CheckSynthesisResult(result, sdk.ResultReason.SynthesizingAudioCompleted);
+            expect(result.audioData.byteLength).toBeGreaterThan(64 << 7); // longer than 1s
```
glharper, line 660: General question about this test: is there something different about what is returned when openRange is set that can be checked?

src/common.speech/SynthesisContext.ts
yulin-li: agree, will update, thx

tests/SpeechSynthesisTests.ts
yulin-li: If `openRange` is set, the `synthesis.context` will contain `"autoDetection":true`. And, the synthesized text is Chinese, if no auto detection, the default english voices cannot read Chinese, then a empty result will be returned.

PR #196, opened 2020-06-30
Add a cache for OCSP responses.

rhurey:
Extend OCSP certificate verification to include a memory and disk based cache.

The disk cache will by default be located in TMPDIR/<username>/ but can be overridden.

When the cached OCSP response is within 1 day of expiring, or 1/2 way through it's validity period (whichever is shorter) a background task is queued to refresh it while the cached value is used.


src/common.browser/CertChecks.ts
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    // Test hook to force the disk cache to be recreated.
+    public static forceReinitDiskCache(): void {
+        CertCheckAgent.privDiskCache = undefined;
+        CertCheckAgent.privMemCache = {};
+    }
+
+    public GetAgent(disableStapling?: boolean): http.Agent {
+        const agent: any = new Agent.Agent(this.CreaetConnectionReturn);
```
glharper, line 70: typo

PR #198, opened 2020-07-02
Switch SDK to use Native promises.

rhurey:
We'll discuss on Monday if this should be a 1.13 or a 1.14 commit.

This converts the SDK to use native promises internally and drops the homemade promise objects.

The SDK surface remains callback based, but this primes a future switch to exposing Promises.

Breaking change wise, some SDK methods that had looked synchronous but in reality started background processes and orphaned them now take callbacks for completion.

One area of work left is to review all empty catch blocks to ensure that eating the error is really our *only* option.


src/sdk/Transcription/ConversationTranslator.ts
```diff
@@ -194,7 +194,7 @@ export class ConversationTranslator implements IConversationTranslator, IDisposa
         try {
 
             // stop the speech websocket
-            this.cancelSpeech();
+            this.cancelSpeech().catch();
```
rhurey, line 197: >.catch(); [](start = 31, length = 9)

Review

src/sdk/Transcription/ConversationTranslator.ts
```diff
@@ -264,7 +264,7 @@ export class ConversationTranslator implements IConversationTranslator, IDisposa
 
                                 this.privIsSpeaking = false;
                                 // this.fireCancelEvent(error);
-                                this.cancelSpeech();
+                                this.cancelSpeech().catch();
```
rhurey, line 267: >.catch(); [](start = 51, length = 9)

Review

src/sdk/Transcription/ConversationTranslator.ts
```diff
@@ -280,7 +280,7 @@ export class ConversationTranslator implements IConversationTranslator, IDisposa
                     ((error: any) => {
                         this.privIsSpeaking = false;
                         // this.fireCancelEvent(error);
-                        this.cancelSpeech();
+                        this.cancelSpeech().catch();
```
rhurey, line 283: >.catch(); [](start = 43, length = 9)

Review

src/sdk/Transcription/ConversationTranslator.ts
```diff
@@ -290,7 +290,7 @@ export class ConversationTranslator implements IConversationTranslator, IDisposa
 
             this.handleError(error, err);
 
-            this.cancelSpeech();
+            this.cancelSpeech().catch();
```
rhurey, line 293: >.catch(); [](start = 31, length = 9)

Review

src/sdk/Audio/SpeakerAudioDestination.ts
```diff
@@ -97,16 +126,16 @@ export class SpeakerAudioDestination implements IAudioDestination, IPlayer {
                     this.privMediaSource.duration = MediaDurationPlaceholderSeconds;
                     this.privSourceBuffer = this.privMediaSource.addSourceBuffer(mimeType);
                     this.privSourceBuffer.onupdate = (_: Event) => {
-                        this.updateSourceBuffer();
+                        this.updateSourceBuffer().catch();
```
rhurey, line 129: >(); [](start = 55, length = 3)

Empty Catch x3

src/sdk/DialogServiceConnector.ts
```diff
@@ -96,8 +96,8 @@ export class DialogServiceConnector extends Recognizer {
      * @function
      * @public
      */
-    public connect(): void {
-        this.privReco.connect();
+    public async connect(): Promise<void> {
```
rhurey, line 99: >connect [](start = 17, length = 7)

Should become callbacks.

src/sdk/DialogServiceConnector.ts
```diff
@@ -173,36 +181,43 @@ export class DialogServiceConnector extends Recognizer {
...
-                        /* tslint:disable:no-empty */
-                    }).on((_: boolean): void => { },
-                        (error: string): void => {
-                            if (!!err) {
-                                err(error);
-                            }
-                        });
+                    }
+                };
+                callbackHolder().catch(() => { });
```
rhurey, line 220: Review the error paths

rhurey:
1.14 is the target.

src/sdk/Audio/SpeakerAudioDestination.ts
yulin-li: I'm ok with this.

src/common/Promise.ts
```diff
@@ -455,3 +262,40 @@ export class Sink<T> {
         this.privSuccessHandlers = [];
     }
 }
+
+export function marshalProimseToCallbacks<T>(
```
glharper, line 266: typo

src/common.speech/ServiceRecognizerBase.ts
```diff
@@ -404,119 +344,105 @@ export abstract class ServiceRecognizerBase implements IDisposable {
...
+            }
 
-                    if (this.receiveMessageOverride !== undefined) {
-                        return this.receiveMessageOverride();
-                    }
-                    if (this.privIsDisposed) {
-                        // We're done.
-                        return PromiseHelper.fromResult(undefined);
-                    }
+            let connection = await this.fetchConnection();
```
yulin-li, line 356: Not for this PR. In `receiveMessage()`, should we start a new connection? The `fetchConnection()` will try to open a new connection if the current connection is closed (mostly due to an error). But the re-opened connection will return no message as we never send any request in this connection.

src/common/Promise.ts
```diff
@@ -105,253 +83,82 @@ export class PromiseResultEventSource<T>  {
...
+                this.privIsCompleted = true;
+                this.privIsError = true;
+                if (!!after) {
+                    after();
                 }
-            },
-        );
-
-        return continuationDeferral.promise();
+                resolve(this);
```
ralph-msft: >resolve(this); [](start = 16, length = 14)

Should this be reject(reason_for_failure)?

src/sdk/Transcription/ConversationTranslator.ts
ralph-msft: Isn't there supposed to be a function handler here for the error cases?

---
In reply to: [449317700](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/198#discussion_r449317700) [](ancestors = 449317700)

src/common.browser/WebsocketConnection.ts
```diff
@@ -79,11 +78,11 @@ export class WebsocketConnection implements IConnection {
             headers);
     }
 
-    public dispose = (): void => {
+    public async dispose(): Promise<void> {
```
ralph-msft, line 81: >public async dispose(): Promise<void> { [](start = 4, length = 39)

I think we've inherited some class designs from the C++ code here in a way that doesn't make sense for JS. We don't really have any native resources to "dispose" of here and I think a cleaner design would be to rename this to close() which makes more sense as async function.

ralph-msft:
>     public disconnect(cb?: () => void, err?: (e: string) => void): void {

Have you considered using your marshalProimseToCallbacks function here?

---
Refers to: src/common.speech/Transcription/ConversationTranslatorRecognizer.ts:130 in 73b39f3. [](commit_id = 73b39f37da6be64107c1a106dcf4208ff1b8a48f, deletion_comment = False)

src/sdk/Transcription/ConversationTranslator.ts
ralph-msft: Ideally if we wanted to handle errors fully and properly here, we would append the errors from the cancelSpeech() operation (line 283) to the error returned in the call to handleError() at line 285 as an "inner error". Alternatively we may want to simply log the inner error and and not report it to the user.

Either way I think not waiting for the cancelSpeech() async operation to complete before invoking the callback may be problematic as we may return to the user in some transient internal state.

---
In reply to: [449317811](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/198#discussion_r449317811) [](ancestors = 449317811)

src/sdk/Transcription/ConversationTranslator.ts
rhurey: Yes. I reviewed all the empty catch() statements and added something. They were initially there to help w/ the linter, and to serve as easily findable review targets.

---
In reply to: [457662561](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/198#discussion_r457662561) [](ancestors = 457662561,449317700)

rhurey:
>     public disconnect(cb?: () => void, err?: (e: string) => void): void {

For this class I'd more likely convert everything to use Promises and push the translation up to the SDK layer.

Maybe an extra follow up PR. Need to unblock some work for Glenn and give him promises to work with.

---
In reply to: [661317657](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/198#issuecomment-661317657) [](ancestors = 661317657)

---
Refers to: src/common.speech/Transcription/ConversationTranslatorRecognizer.ts:130 in 73b39f3. [](commit_id = 73b39f37da6be64107c1a106dcf4208ff1b8a48f, deletion_comment = False)

src/common.browser/WebsocketConnection.ts
rhurey: While I don't disagree, next time.

---
In reply to: [457669739](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/198#discussion_r457669739) [](ancestors = 457669739)

src/common.speech/ServiceRecognizerBase.ts
rhurey: It's arguable. If audio is flowing (or should be, maybe it's suspended in the wait throttle) one or the other will open the connection. If message flow was stopping, I expect the queue drain from the old connection to have stopped it.


---
In reply to: [456227814](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/198#discussion_r456227814) [](ancestors = 456227814)

src/common/Promise.ts
```diff
@@ -104,254 +82,30 @@ export class PromiseResultEventSource<T>  {
...
-
-        return this.on(callbackWrapper, callbackWrapper);
-    }
-}
-
 // tslint:disable-next-line:max-classes-per-file
 export class Deferred<T> implements IDeferred<T> {
     private privPromise: Promise<T>;
-    private privSink: Sink<T>;
+    private privResolve: (value?: T | PromiseLike<T>) => void;
```
modnarb, line 88: From what I'm reading, the "Like" types are for broad implementation compatibility. But why the use of PromiseLike here but not below on the 'resolve'?

    public resolve = (result: T | Promise<T>): Deferred<T> => {

PR #205, opened 2020-07-10
Parse json before sending message payload

modnarb:


rhurey:
Is the goal here to validate the JSON is valid before transmission?

If so, there should be a test of SendActivity that tries to send invalid JSON.

modnarb:
Added a test to cover invalid JSON

tests/DialogServiceConnectorTests.ts
```diff
@@ -981,7 +957,7 @@ describe.each([true, false])("Service-based tests", (forceNodeWebSocket: boolean
 
         connector.speechStartDetected = (sender: sdk.DialogServiceConnector, e: sdk.RecognitionEventArgs): void => {
             try {
-                const message: any = { speak: "This is speech", text: "This is text", type: "message" };
+                const message: string = '{ "speak" : "This is speech", "text": "This is text", "type" : "message" }';
```
glecaros, line 960: nit: maybe use an object and then stringify...

PR #210, opened 2020-07-15
Reconnect with conversation IDs

modnarb:


src/sdk/DialogServiceConnector.ts
```diff
@@ -271,7 +271,7 @@ export class DialogServiceConnector extends Recognizer {
                 commType: communicationType,
                 commandsCulture: undefined,
                 connectionId: this.properties.getProperty(PropertyId.Conversation_ApplicationId),
-                conversationId: undefined,
+                conversationId: this.privProperties.getProperty(PropertyId.Conversation_Conversation_Id, undefined),
```
encorona-ms, line 274: nit: you can use this.properties instead

PR #213, opened 2020-07-16
get safari-friendly audioContext instance

glharper:
... add playResultAudio method to BaseAudioPlayer
Addresses issue "Safari fallback to webkitAudioContext is done in the sample app instead of the SDK."

### DIFF
--- a/src/sdk/Audio/AudioStreamFormat.ts
+++ b/src/sdk/Audio/AudioStreamFormat.ts
@@ -19,6 +19,31 @@ export abstract class AudioStreamFormat {
     }
 
     /**
+     * Creates an audio context appropriate to current browser
+     * @member AudioStreamFormat.getAudioContext
+     * @function
+     * @public
+     * @returns {AudioContext} An audio context instance
+     */
+    public static getAudioContext(sampleRate?: number): AudioContext {
+        // Workaround for Speech SDK bug in Safari.
+        const AudioContext = (window as any).AudioContext // our preferred impl
+            || (window as any).webkitAudioContext // fallback, mostly when on Safari
+            || false; // could not find.
+
+        // https://developer.mozilla.org/en-US/docs/Web/API/AudioContext
+        if (!!AudioContext) {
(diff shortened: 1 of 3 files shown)


src/sdk/Audio/AudioStreamFormat.ts
```diff
@@ -18,6 +18,31 @@ export abstract class AudioStreamFormat {
         return AudioStreamFormatImpl.getDefaultInputFormat();
     }
 
+    /**
+     * Creates an audio context appropriate to current browser
+     * @member AudioStreamFormat.getAudioContext
+     * @function
+     * @public
+     * @returns {AudioContext} An audio context instance
```
rhurey, line 26: This should move to the Impl class to not have it published on the SDK surface.

src/sdk/Audio/BaseAudioPlayer.ts
```diff
@@ -23,11 +23,35 @@ export class BaseAudioPlayer {
...
     }
 
+    /**
+     * play Audio from result using safari-friendly audioContext methods
+     * @member BaseAudioPlayer.prototype.playResultAudio
+     * @function
+     * @public
+     * @param {ArrayBuffer} audioData audio data to be played.
+     */
+    public playResultAudio(audioData: ArrayBuffer): void {
```
rhurey, line 42: >playResultAudio [](start = 11, length = 15)

Why do we need a new public API for this?

glharper: The clearBuffer() call in updateAudioBuffer on line 135 throws in Safari. I'm not sure how playAudioSample is currently being used in customer code, but will look and see if i can figure out the issue.

src/sdk/Audio/AudioStreamFormat.ts
glharper: Done.

src/sdk/Audio/BaseAudioPlayer.ts
glharper: Moved this to private method, public method will call this when webkit detected.

PR #231, opened 2020-08-26
Reconnect when timeout is honored by backend

glharper:
…backend. This is intended to spark a discussion and crystallize the correct approach.


glharper:
This code fixes the translation reco long running disconnect test. For transcriber disconnect, it works if maxConnectionDurationSecs is set to 12s (for a 30s wav file), but connection duration of 10s produces a "turn started before last turn end" error in the test.

tests/ConversationTranscriberTests.ts
```diff
@@ -386,3 +386,63 @@ test("Create Conversation with one channel audio (aligned)", (done: jest.DoneCal
             done.fail(error);
         });
 });
+test("Create Conversation and force disconnect", (done: jest.DoneCallback) => {
+    // tslint:disable-next-line:no-console
+    console.info("Name: Create Conversation and force disconnect");
+    const s: sdk.SpeechTranslationConfig = BuildSpeechConfig();
+    s.setServiceProperty("maxConnectionDurationSecs", "12", sdk.ServicePropertyChannel.UriQueryParameter);
```
brandom-msft, line 393: 12 is used because the timeout is 10s on the backend, yes? Might be worth adding a comment why for future people

brandom-msft (approved):

tests/ConversationTranscriberTests.ts
glharper: Added.

PR #241, opened 2020-09-04
Updates to ID usage, parity with native SDK

brandom-msft:
Addressing a few issues, I think the first two might be due to overloaded uses of the idea of "session".

In native, for DialogServiceConnector, a turn interaction (like SessionStarted/Stopped) uses the interactionId of the latest context message
https://msasg.visualstudio.com/Skyman/_git/Carbon?path=%2Fsource%2Fcore%2Fsr%2Faudio_stream_session.cpp&version=GBmaster&line=1162&lineEnd=1163&lineStartColumn=1&lineEndColumn=1&lineStyle=plain

Also in native, sessionId is used as the connectionId to correlate logs:
https://msasg.visualstudio.com/Skyman/_git/Carbon?path=%2Fsource%2Fcore%2Fsr%2Fusp_reco_engine_adapter.cpp&version=GBmaster&line=345&lineEnd=347&lineStartColumn=1&lineEndColumn=112&lineStyle=plain

During connection setup the query parameter "connectionId" is not set, instead X-ConnectionId is set which should only be a header param

### DIFF
--- a/tests/ConversationTranscriberTests.ts
+++ b/tests/ConversationTranscriberTests.ts
@@ -386,3 +386,64 @@ test("Create Conversation with one channel audio (aligned)", (done: jest.DoneCal
             done.fail(error);
         });
 });
+test("Create Conversation and force disconnect", (done: jest.DoneCallback) => {
+    // tslint:disable-next-line:no-console
+    console.info("Name: Create Conversation and force disconnect");
+    const s: sdk.SpeechTranslationConfig = BuildSpeechConfig();
+    // Use 12s timeout since backend timeout is 10s.
+    s.setServiceProperty("maxConnectionDurationSecs", "12", sdk.ServicePropertyChannel.UriQueryParameter);
+    objsToClose.push(s);
+
+    const c: sdk.Conversation = CreateConversation(s);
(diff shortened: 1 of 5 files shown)


brandom-msft:
Seeing some extra websocket connections hanging out during local tests, need to follow-up on this behavior

tests/ConversationTranscriberTests.ts
```diff
@@ -386,3 +386,64 @@ test("Create Conversation with one channel audio (aligned)", (done: jest.DoneCal
             done.fail(error);
         });
 });
+test("Create Conversation and force disconnect", (done: jest.DoneCallback) => {
```
brandom-msft, line 389: Not sure why this and the other Transcription change are present here.. I did a rebase from master which I expected would resolve this, but these are not actually changes I've made

glharper (approved):

tests/ConversationTranscriberTests.ts
glharper: This and some of the ServiceRecognizerBase changes were in my merged PR for the reconnect-fix branch.

PR #244, opened 2020-09-22
Fix for InitialSilenceTimeout halting subsequent ListenOnce calls

brandom-msft:
With the previous logic, on any result returned with a status other than "Success", the 'recognized' event is never fired and blocks any further actions on the recognizer. Instead we'll allow for all non-error status to return in the expected flow.


glharper (approved):

ralph-msft (approved):
:shipit:

PR #257, opened 2020-10-12
Add CELA requested text blob to LICENSE and NOTICE files

glharper:
Requested by Oliver


rhurey:
#259 will require some more additions to this about UUID.

brandom-msft (approved):

Issue #242, opened 2020-09-09
CPU is exploding on load


bivas6:
Hi,

We are trying to run multiple simultaneously `startContinuousRecognitionAsync` requests, but CPU is exploding.

We ran about 7 requests/second and in about ~3 minutes, the machine is crashed. 
Note that we are able to make 20 requests/second with [this package](https://github.com/orgads/ms-bing-speech-service#readme), and CPU is about 40-50% usage.

chrome inspector cpuprofile file is attached. 
[azure-ms-cpu.zip](https://github.com/microsoft/cognitive-services-speech-sdk-js/files/5195385/azure-ms-cpu.zip)

cc: @orgads 

Thanks!

trrwilson:
Thanks for getting in touch!

We'll take a look at this apparent leak behavior ASAP and report back. How are you managing audio input for the 7 requests per second? If there's a sample usage you could share, that'd for sure be very helpful.

bivas6:
```
import * as sdk from 'microsoft-cognitiveservices-speech-sdk';
import { Readable } from 'stream';
import getToken from './getToken';
import * as fs from 'fs'
import * as Throttle from 'throttle'
import * as path from 'path';

async function sendStream(stream: Readable) {
  const pushStream = sdk.AudioInputStream.createPushStream();
  const handleData = (buf: Buffer) => pushStream.write(buf.slice().buffer);

  stream
    .on('data', handleData)
    .once('close', () => {
      pushStream.close();
      stream.removeListener('data', handleData);
    });

  const audioConfig = sdk.AudioConfig.fromStreamInput(pushStream);
  const authToken = await getToken(); //getToken create and update valid authrization tokens
  const url = new URL('wss://westus2.stt.speech.microsoft.com/speech/recognition/dictation/cognitiveservices/v1');
  const speechConfig = sdk.SpeechConfig.fromEndpoint(url, '');
  speechConfig.authorizationToken = authToken;
  const recognizer = new sdk.SpeechRecognizer(speechConfig, audioConfig);
  _registerEvents(recognizer);
  recognizer.startContinuousRecognitionAsync(
    undefined,
    (err) => {
      console.error('err - ' + err);
      recognizer.close();
    }
  );
}

function _registerEvents(reco: sdk.SpeechRecognizer) {
  reco.recognizing = (_s: sdk.Recognizer, event: sdk.SpeechRecognitionEventArgs) => {
  };

[...]

PR #258, opened 2020-10-13
Websocket: Disable perMessageDeflate (#242)

orgads:
It affects performance badly.


rhurey:
This strikes me as something that should have a string control property to it so we can enable / disable.

Need to plumb that down to the adapter.

orgads:
Any progress with this?

glharper:
See https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/271

orgads:
Thanks

PR #260, opened 2020-10-16
[Audio Config] - Add support media stream track

enricop89:
Hi, 

I added a new function in the Audio Config file to allow the injection of a MediaTrack. This is useful when you are using the SDK with a third party library for Video (for example, [Vonage](https://tokbox.com/developer/guides/basics/))

### DIFF
--- a/src/common.browser/MicAudioSource.ts
+++ b/src/common.browser/MicAudioSource.ts
@@ -63,8 +63,10 @@ export class MicAudioSource implements IAudioSource {
 
     public constructor(
         private readonly privRecorder: IRecorder,
-        private readonly deviceId?: string,
-        audioSourceId?: string) {
+        private readonly deviceId?: string,        
+        audioSourceId?: string,
+        private readonly mediaStream?: MediaStream
+        ) {
 
         this.privOutputChunkSize = MicAudioSource.AUDIOFORMAT.avgBytesPerSec / 10;
         this.privId = audioSourceId ? audioSourceId : createNoDashGuid();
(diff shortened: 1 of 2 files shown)


src/common.browser/MicAudioSource.ts
```diff
@@ -64,7 +64,9 @@ export class MicAudioSource implements IAudioSource {
     public constructor(
         private readonly privRecorder: IRecorder,
         private readonly deviceId?: string,
-        audioSourceId?: string) {
+        audioSourceId?: string,
+        private readonly mediaStream?: MediaStream
```
rhurey, line 68: >mediaStream [](start = 25, length = 11)

Why create another member variable instead of moving the initialization of the existing privMediaStream?

enricop89: Yes, makes sense. I fixed it.

glharper:
Do we need a new API method for this? Feels like this should go into fromStreamInput, since there's already type checking there.

enricop89:
@glharper you mean like adding a third [option](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/master/src/sdk/Audio/AudioConfig.ts#L91) ?

Something like: 

```
public static fromStreamInput(audioStream: AudioInputStream | PullAudioInputStreamCallback | 
    mediaStream: MediaStream): AudioConfig {}
```

glharper:
@enricop89 Yes, exactly.

enricop89:
@glharper yup, sounds ok. I updated the PR.

glharper (approved):

Issue #268, opened 2020-11-09
AudioInputStream Error: "ReferenceError: Blob is not defined"


yoshigev:
Hi,

We are using SDK version 1.14.0.

We are using AudioInputStream for passing audio to Speaker Verification, but I suspect that the issue is relevant also for other usages.

Whenever we call `pushStream.close()`, we face the following error:
```
ReferenceError: Blob is not defined
```

I suspect that the following lines of code in `AudioInputStream.js`:
```js
                            if (typeof (XMLHttpRequest) !== "undefined") {
                                return Promise.resolve(new Blob(data));
                            }
```
Should actually be:
```js
                            if (typeof (Blob) !== "undefined") {
                                return Promise.resolve(new Blob(data));
                            }
```

Note that this problem probably occurs because somewhere else in our code there is the following line (which we need for other purposes):
```ts
globalAny.XMLHttpRequest = require('xhr2');
```

cc: @orgads,@bivas6

Thanks.

amitkumarshukla:
@glharper Could you please take a look at this one.

glharper:
@yoshigev 1.14.1 has been released with a fix for this issue. Please update and provide feedback if you're so inclined. 

Glenn

yoshigev:
The problem is fixed with the new version. Thank you very much.

PR #269, opened 2020-11-09
GH #268 - check if Blob is defined before return Blob type

glharper:
https://github.com/microsoft/cognitive-services-speech-sdk-js/issues/268

### DIFF
--- a/src/sdk/Audio/AudioInputStream.ts
+++ b/src/sdk/Audio/AudioInputStream.ts
@@ -188,7 +188,7 @@ export class PushAudioInputStreamImpl extends PushAudioInputStream implements IA
             const readCycle = (): Promise<Blob | Buffer> => {
                 return audioNode.read().then<Blob | Buffer>((audioStreamChunk: IStreamChunk<ArrayBuffer>) => {
                     if (!audioStreamChunk || audioStreamChunk.isEnd) {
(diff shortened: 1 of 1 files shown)


brandom-msft (approved):

src/sdk/Audio/AudioInputStream.ts
```diff
@@ -188,7 +188,7 @@ export class PushAudioInputStreamImpl extends PushAudioInputStream implements IA
             const readCycle = (): Promise<Blob | Buffer> => {
                 return audioNode.read().then<Blob | Buffer>((audioStreamChunk: IStreamChunk<ArrayBuffer>) => {
                     if (!audioStreamChunk || audioStreamChunk.isEnd) {
-                        if (typeof (XMLHttpRequest) !== "undefined") {
+                        if (typeof (XMLHttpRequest) !== "undefined" && typeof (Blob) !== "undefined") {
```
yoshigev, line 191: Are you sure that it is necessary to check `XMLHttpRequest` here at all? On line 197 only `Blob` is checked.

glharper: Yes. There is an alternate case in Node where Blob is defined but XMLHttpRequest is not, and a Buffer type must be returned then.

PR #270, opened 2020-11-11
pronunciation assessment implementation

yulin-li:
pronunciation assessment implementation

### DIFF
--- a/tests/PronunciationAssessmentTests.ts
+++ b/tests/PronunciationAssessmentTests.ts
@@ -1,0 +1,172 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as sdk from "../microsoft.cognitiveservices.speech.sdk";
+import {
+    PronunciationAssessmentGradingSystem,
+    PronunciationAssessmentGranularity
+} from "../microsoft.cognitiveservices.speech.sdk";
+import {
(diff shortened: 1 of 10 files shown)


glharper (approved):

tests/PronunciationAssessmentTests.ts
```diff
@@ -0,0 +1,172 @@
...
+    test("test Pronunciation Assessment", (done: jest.DoneCallback) => {
+        // tslint:disable-next-line:no-console
+        console.info("Name: test Pronunciation Assessment");
+        const s: sdk.SpeechConfig = BuildSpeechConfig();
+        objsToClose.push(s);
+
+        const r: sdk.SpeechRecognizer = BuildRecognizerFromWaveFile(s, Settings.WaveFile);
+        objsToClose.push(r);
+
+        const p: sdk.PronunciationAssessmentConfig = new sdk.PronunciationAssessmentConfig("");
```
yinhew, line 137: Can you put some reference text here? empty reference text actually doesn't trigger our score AM on service side.

yinhew (approved):

tests/PronunciationAssessmentTests.ts
yulin-li: as we talked offline. current test region doesn't support pronunciation assessment. will update the test case after pronunciation assessment is enabled on that region.

PR #272, opened 2020-11-12
fix sending speech.config many times in one TTS session

yulin-li:
Currently, the SDK sends `speech.config` before every audio chunk. This PR fixes this issue.

### DIFF
--- a/src/common.speech/SynthesisAdapterBase.ts
+++ b/src/common.speech/SynthesisAdapterBase.ts
@@ -216,26 +216,29 @@ export class SynthesisAdapterBase implements IDisposable {
         this.privErrorCallback = errorCallBack;
 
         this.privSynthesisTurn.startNewSynthesis(requestId, text, isSSML, audioDestination);
+
         try {
-            const connection: IConnection = await this.fetchConnection();
-            await this.sendSynthesisContext(connection);
-            await this.sendSsmlMessage(connection, ssml, requestId);
-            const synthesisStartEventArgs: SpeechSynthesisEventArgs = new SpeechSynthesisEventArgs(
-                new SpeechSynthesisResult(
-                    requestId,
-                    ResultReason.SynthesizingAudioStarted,
-                )
-            );
-
-            if (!!this.privSpeechSynthesizer.synthesisStarted) {
-                this.privSpeechSynthesizer.synthesisStarted(this.privSpeechSynthesizer, synthesisStartEventArgs);
-            }
-
-            const messageRetrievalPromise = this.receiveMessage();
+            await this.connectImpl();
         } catch (e) {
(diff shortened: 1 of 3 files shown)


src/common.speech/SynthesisAdapterBase.ts
```diff
@@ -146,9 +144,11 @@ export class SynthesisAdapterBase implements IDisposable {
         this.connectionEvents.attach((connectionEvent: ConnectionEvent): void => {
             if (connectionEvent.name === "ConnectionClosedEvent") {
                 const connectionClosedEvent = connectionEvent as ConnectionClosedEvent;
-                this.cancelSynthesisLocal(CancellationReason.Error,
-                    connectionClosedEvent.statusCode === 1007 ? CancellationErrorCode.BadRequestParameters : CancellationErrorCode.ConnectionFailure,
-                    connectionClosedEvent.reason + " websocket error code: " + connectionClosedEvent.statusCode);
+                if (connectionClosedEvent.statusCode !== 1000) {
```
glharper, line 147: What happens in this flow if the statusCode 1000 is received? We silently ignore the closed event?

src/common.speech/SynthesisAdapterBase.ts
```diff
@@ -216,26 +216,29 @@ export class SynthesisAdapterBase implements IDisposable {
...
 
-            const messageRetrievalPromise = this.receiveMessage();
+        try {
+            await this.connectImpl();
         } catch (e) {
             this.cancelSynthesisLocal(CancellationReason.Error, CancellationErrorCode.ConnectionFailure, e);
             return Promise.reject(e);
         }
+
+        const connection: IConnection = await this.fetchConnection();
```
glharper, line 227: So this line and the block below are no longer in a try/catch block. What happens if one of the lines throws an exception now?

glharper (commented):
See comments

src/common.speech/SynthesisAdapterBase.ts
yulin-li: Thanks for reviewing. `statusCode` 1000 means normal closure, which is normally caused by client close the connection. The error code info could be found in our [internal wiki ](https://speechwiki.azurewebsites.net/partners/protocol-websockets-tts.html?q=websocket)

src/common.speech/SynthesisAdapterBase.ts
yulin-li: thx, I will move this into try/catch block.

glharper (approved):
Approved with addition of try/catch block

yinhew (commented):
Is this issue JS specific?

yulin-li:
@yinhew yes, it's JS specific.

yinhew (approved):

PR #283, opened 2020-12-03
Fix issue 881 for re-using recognizer object

glharper:
https://github.com/Azure-Samples/cognitive-services-speech-sdk/issues/881


brandom-msft:
So why was the failure manifested without the ".active" check? Something getting mangled in between turns?

brandom-msft (approved):

glharper:
@brandom-msft That privMediaStream member gets saved as a member variable now as part of [this PR for reusing an existing MediaStream](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/260), and re-using the audioConfig instance maintains the reference to the old now-inactive MediaStream.

PR #285, opened 2020-12-08
Add node-only tests, update file read to accomodate node

glharper:
On the node side, there is an issue where PushInputAudioStream does not do wav header checking like FileAudioSource, and there may be performance issues as well. For now, tests affected have been conditionally disabled for Node test runs, as have browser specific tests.


yulin-li:
:shipit:

yulin-li (approved):
:shipit:

PR #294, opened 2020-12-16
Add AudioConfig.fromReadableStream implementation

SpaceKatt:
This PR adds a static method on `AudioConfig` to create an `AudioConfig` object from a `Readable` node stream.

## Why this PR?

I ran into an issue where it was difficult to use this SDK while streaming audio over UDP into a Docker container. To solve the problem at hand, I had to read the source code of the SDK and figure out how the SDK internally handled audio streams. 

The code in this PR enabled me to use this SDK while streaming audio over UDP into a Docker container. This method will hopefully save future devs time when they try to implement similar use cases.

## Example `ffmpeg` UDP stream command

`/usr/bin/ffmpeg -guess_layout_max 0 -ac 1 -f alsa -i hw:CARD=<INSERT_AUDIO_DEVICE_HERE> -ar 44100 -f s16le -acodec pcm_s16le -loglevel warning -nostats udp://127.0.0.1:1337`

## Example UDP client code

This snippet is modified from a working solution to demonstrate the possible use of `AudioConfig.fromReadableStream` (and obviously doesn't compile without modification).

```ts
import { AudioConfig } from 'microsoft-cognitiveservices-speech-sdk'
import { createSocket } from 'dgram';
import { PassThrough } from 'stream'

const port = 1337;
const audioStream = new PassThrough();

const socket = createSocket('udp4');
socket.on('listening', () => {
    console.log(`UDP input listening on port ${port}`)
});
socket.on('message', (msg: Buffer) => {
    audioStream.push(msg);
});
socket.bind(port);

const audioConfig = AudioConfig.fromReadableStream(audioStream);

// ...use audioConfig to build SpeechRecognizer
```

### DIFF
--- a/src/sdk/Audio/AudioConfig.ts
+++ b/src/sdk/Audio/AudioConfig.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 import { PathLike } from "fs";
+import { Readable } from "stream";
 import {
     FileAudioSource,
     MicAudioSource,
@@ -103,6 +104,28 @@ export abstract class AudioConfig {
         }
 
         throw new Error("Not Supported Type");
+    }
+
+    /**
+     * Creates an AudioConfig object from a standard Readable stream.
+     * Useful when streaming audio from external network sources (e.g., ffmpeg UDP stream into a Docker container).
+     * @member AudioConfig.fromReadableStream
+     * @function
+     * @public
+     * @param {Readable} audioStream - Custom audio input from a Readable stream. Currently, only WAV / PCM is supported.
+     * @param {AudioStreamFormat} format - The audio data format in which audio
+     *        will be returned from the stream's read() method.
+     * @returns {AudioConfig} The audio output configuration being created.
+     */
+    public static fromReadableStream(audioStream: Readable, format?: AudioStreamFormat): AudioConfig {
+        const pushStream = AudioInputStream.createPushStream(format);
+
+        audioStream.on('readable', () => {
+            const buffer = audioStream.read();
+            pushStream.write(buffer);
+        });
+
+        return AudioConfig.fromStreamInput(pushStream);
     }
 
     /**


glharper:
@SpaceKatt Thank you for taking the time to create this PR with an explanation of your reasoning and the specific use case, very much appreciated!

A couple of questions:
- Given this library's support of both Node and browser users, how exactly should browser users consume this, given the Node stream dependency? 
- How do you view the tradeoffs of adding a new API with a new dependency vs. documenting the five lines of code for binding to Readable streams?

Thanks,
Glenn

SpaceKatt:
> @SpaceKatt Thank you for taking the time to create this PR with an explanation of your reasoning and the specific use case, very much appreciated!
> 
> A couple of questions:
> 
> * Given this library's support of both Node and browser users, how exactly should browser users consume this, given the Node stream dependency?
> * How do you view the tradeoffs of adding a new API with a new dependency vs. documenting the five lines of code for binding to Readable streams?
> 
> Thanks,
> Glenn

Hmm... those are some darn good questions! I had not considered browser users when creating this PR. Given the bowser support requirement, I no longer think this PR is tenable.

Is there documentation where I could contribute to document this methodology? If so, then please link it and I'll open a PR there. `n_____n`

Feel free to close this PR, be well to yourself and have fun!

src/sdk/Audio/AudioConfig.ts
```diff
@@ -105,6 +106,28 @@ export abstract class AudioConfig {
...
+     * @returns {AudioConfig} The audio output configuration being created.
+     */
+    public static fromReadableStream(audioStream: Readable, format?: AudioStreamFormat): AudioConfig {
+        const pushStream = AudioInputStream.createPushStream(format);
+
+        audioStream.on('readable', () => {
+            const buffer = audioStream.read();
+            pushStream.write(buffer);
+        });
+
```
rhurey, line 127: Nothing ever closes the push stream. That can lead to recognition hanging when the Readable runs out of data.

SpaceKatt: Excellent point! I've updated the code to close the push stream when `read()` returns `null`.

Are there any other cases you believe should be handled? (e.g., do you have thoughts around `.on('error', () => { ??? })`)

PR #295, opened 2020-12-18
Glharper/file audio node

glharper:
(No squash when merging)
Allow FileAudioSource and upstream fromWavFileInput to take a Buffer input, making Node usage of fromWavFileInput possible. (Also some test connection leakage fixes.)
Open questions: 
- Should FileAudioSource be renamed?
- Should it be moved to common (instead of staying in common.browser)?


dargilco (approved):

src/common.speech/ServiceRecognizerBase.ts
```diff
@@ -214,6 +211,8 @@ export abstract class ServiceRecognizerBase implements IDisposable {
 
         this.privRecognizerConfig.SpeechServiceConfig.Context.audio = { source: deviceInfo };
 
+        // Start the connection to the service. The promise this will create is stored and will be used by configureConnection().
+        const conPromise: Promise<IConnection> = this.connectImpl();
```
rhurey, line 215: The connection was being opened before the audio was started intentionally to not serialize the two long running operations. Running them in series imparts a multi-hundred millisecond penalty.

src/sdk/TranslationRecognizer.ts
```diff
@@ -206,7 +206,7 @@ export class TranslationRecognizer extends Recognizer {
      * @param cb - Callback that received the translation has stopped.
      * @param err - Callback invoked in case of an error.
      */
-    public stopContinuousRecognitionAsync(cb?: () => void, err?: (e: string) => void): void {
+    public stopContinuousRecognitionAsync(cb?: (v: unknown) => void, err?: (e: string) => void): void {
```
rhurey, line 209: >v: unknown [](start = 48, length = 10)

Why does the stop callback now take a  parameter?

package.json
```diff
@@ -66,16 +66,16 @@
...
     "ocsp": "^1.2.0",
     "request": "^2.88.0",
     "rimraf": "^3.0.2",
     "semver": "^6.3.0",
     "source-map-loader": "^0.2.4",
-    "ts-jest": "^25.5.1",
+    "ts-jest": "^26.4.4",
     "tslint": "^5.20.1",
-    "typescript": "^3.5.3",
+    "typescript": "^3.8",
```
rhurey, line 78: >"typescript": "^3.8", [](start = 4, length = 21)

Are these two minor releases .d.ts file compatable?

src/common.speech/ServiceRecognizerBase.ts
glharper: So should I restore the original order and wrap the audio start in a try/catch which closes (or undoes establishing steps for) the connection on error?

package.json
glharper: How can I check whether it is or not?

src/sdk/TranslationRecognizer.ts
glharper: this is called on line 276 of ConversationTranslator.stopTranscribingAsync(). VS Code was complaining about:
```
                await new Promise((resolve: () => void, reject: (error: string) => void): void => {
                    this.privTranslationRecognizer?.stopContinuousRecognitionAsync(resolve, reject);
                });
```
It wanted resolve to have a "value: unknown" param, and this TranslationRecognizer change is downstream of that.

src/common.speech/ServiceRecognizerBase.ts
glharper: Refactored as outlined in my previous comment.

src/sdk/TranslationRecognizer.ts
glharper: reverted

package.json
glharper: reverted

rhurey (approved):

PR #301, opened 2021-01-06
DialogServiceConnector: add a new 'TurnStatusReceived' event 

trrwilson:
## The problem

Dialog interactions with a Bot can fail for reasons that also make communicating the failure impossible: the destination host being unreachable because of a network routing issue, a business logic operation on the bot (e.g. DB lookup) exceeding a maximum timeout and getting "cut off," an unhandled exception in the bot itself, and so on. Since this is the connection between Direct Line Speech and the bot and not anything direct to the device, in these circumstances, no information whatsoever is communicated back to the client about a failure occurring. That's problematic when you want to support proper reprompt/retry behavior.

## The solution

The Direct Line Speech service now supports per-turn (as defined by a bot's turn contexts) status message, which can inform the client of failures that happen on the other end of the Device<->DLS communication. This change exposes those messages as a new `TurnStatusReceived` event (peer to `ActivityReceived`) and by default opts in to producing them. If for any reason they're problematic, they can also be disabled via a property.

## Example use

```javascript
connector.activityReceived = (sender, eventArgs) => {
  // Do stuff with activities like usual
};
connector.turnStatusReceived = (sender, eventArgs) => {
  // Status codes reflect standard HTTP values like "OK" (200) and "Bad request" (400)
  if (eventArgs.statusCode != 200) {
    promptForRetry(eventArgs.conversationId);
  }
}
```

## Opting out

```javascript
var botConfig = SpeechSDK.BotFrameworkConfig.fromSubscription(
  subscriptionKey,
  serviceRegion
);

botConfig.setProperty(propertyId.Conversation_Request_Bot_Status_Messages, "false");

var connector = new SpeechSDK.DialogServiceConector(botConfig, audioConfig);
...
```


src/common.speech/DialogServiceAdapter.ts
```diff
@@ -208,28 +209,8 @@ export class DialogServiceAdapter extends ServiceRecognizerBase {
...
-                    const activityPayload: ActivityPayloadResponse = ActivityPayloadResponse.fromJSON(connectionMessage.textBody);
-                    const turn = this.privTurnStateManager.GetTurn(responseRequestId);
-
-                    // update the conversation Id
-                    if (activityPayload.conversationId) {
-                        const updateAgentConfig = this.agentConfig.get();
-                        updateAgentConfig.botInfo.conversationId = activityPayload.conversationId;
-                        this.agentConfig.set(updateAgentConfig);
-                    }
+                    this.handleResponseMessage(connectionMessage);
```
glharper, line 212: Nice modularization

src/common.speech/DialogServiceAdapter.ts
```diff
@@ -585,4 +566,57 @@ export class DialogServiceAdapter extends ServiceRecognizerBase {
...
+            case "messagestatus":
+                if (!!this.privDialogServiceConnector.turnStatusReceived) {
+                    try {
+                        this.privDialogServiceConnector.turnStatusReceived(
+                            this.privDialogServiceConnector,
+                            new TurnStatusReceivedEventArgs(responseMessage.textBody));
+                        /* tslint:disable:no-empty */
+                    } catch (error) {
+                        // Not going to let errors in the event handler
+                        // trip things up.
```
glharper, line 611: Why do we do this? Is there some way to give an error some form of logging and then continue on our way?

src/common.speech/ServiceMessages/TurnStatusPayload.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    public get conversationId(): string {
+        return this.privMessageStatusResponse.conversationId;
+    }
+
+    public get statusCode(): any {
+        // Payloads may contain a limited set of textual representations or a numeric status
+        // code. The textual values are here converted into numeric ones.
+        switch (this.privMessageStatusResponse.statusCode) {
```
glharper, line 32: This is probably a better experience for the customer than "guess which random string the backend is using to indicate a time out," but it still feels odd returning a number to clearly indicate the status.

glharper (approved):

src/common.speech/DialogServiceAdapter.ts
trrwilson: That's a good question--do we have an "error trace" equivalent here? I don't see one in use but the idea is provide customer diagnosis assistance if an unexpected code is coming back. If we have something better than a lint-dodging console.log, I'd love to use it.

src/common.speech/ServiceMessages/TurnStatusPayload.ts
trrwilson: The good news is that this is, for the time being, exclusively going to be standard HTTP codes, e.g. 200 for success, 400 for failure (Bad Request), 429 for Timeout (Bot Framework "too many requests" throttling), etc.. If there were a strongly-typed way to represent it (like "HttpStatusCode.OK"), I'd use it, but it seems like we'd have to define that ourselves--likely more harmful than helpful.

src/common.speech/DialogServiceAdapter.ts
glharper: Perhaps emitting a BackgroundEvent would work here:
```
Events.instance.onEvent(new BackgroundEvent(error));
```

glharper (commented):
Should we add a test for this?

src/common.speech/DialogServiceAdapter.ts
trrwilson: Aha, that's exactly what I was looking for! And it's even used in the same file; shame on me for not seeing it. Looked for "log," "trace," and other variations, but never thought to look for "event!"

trrwilson:
> Should we add a test for this?

Absolutely; the "unhappy path" bot behavior will come in a bit later, but I'll add the "happy path" to the existing tests now.

tests/DialogServiceConnectorTests.ts
```diff
@@ -312,6 +313,15 @@ test("ListenOnceAsync", (done: jest.DoneCallback) => {
         }
     };
 
+    connector.turnStatusReceived = (sender: sdk.DialogServiceConnector, e: sdk.TurnStatusReceivedEventArgs) => {
+        turnStatusCounter++;
+        try {
+            expect(e.statusCode == 200);
```
glharper, line 319:
```
            expect(e.statusCode === 200);
```

tests/DialogServiceConnectorTests.ts
```diff
@@ -329,6 +339,7 @@ test("ListenOnceAsync", (done: jest.DoneCallback) => {
         });
 
     WaitForCondition(() => (recoCounter === 2), done);
+    WaitForCondition(() => (turnStatusCounter == 1), done);
```
glharper, line 342:
```
WaitForCondition(() => (turnStatusCounter === 1), done);
```

src/common.speech/DialogServiceAdapter.ts
```diff
@@ -614,8 +614,8 @@ export class DialogServiceAdapter extends ServiceRecognizerBase {
                 break;
 
             default:
-                // tslint:disable-next-line:no-console
-                console.log("Unexpected response of type " + responsePayload.messageType + ". Ignoring.");
+                Events.instance.onEvent(
+                    new BackgroundEvent("Unexpected response of type '" + responsePayload.messageType + "'. Ignoring."));
```
glharper, line 618: A bit pedantic, but I like the newer backtick interpolated string literals, ala:
```
new BackgroundEvent(`Unexpected response of type '${responsePayload.messageType}'. Ignoring.`));
```

tests/DialogServiceConnectorTests.ts
trrwilson: Very astute catch -- manual re-run of the test got it, too:
```
2021-01-12T18:21:07.6639690Z ERROR: D:/a/1/s/tests/DialogServiceConnectorTests.ts:319:33 - == should be ===
2021-01-12T18:21:07.6641862Z ERROR: D:/a/1/s/tests/DialogServiceConnectorTests.ts:342:47 - == should be ===
```

I hadn't configured local testing yet as I wasn't sure how to do it. It's not particularly easy without the pipeline automating secrets (and I don't think everyone will have all the things available/configured), but I'll add a blurb to the readme to help future contributors, too.

tests/DialogServiceConnectorTests.ts
trrwilson: Very astute catch -- manual re-run of the test pipeline after pushing got it, too:
```
2021-01-12T18:21:07.6639690Z ERROR: D:/a/1/s/tests/DialogServiceConnectorTests.ts:319:33 - == should be ===
2021-01-12T18:21:07.6641862Z ERROR: D:/a/1/s/tests/DialogServiceConnectorTests.ts:342:47 - == should be ===
```

I hadn't configured local testing yet as I wasn't sure how to do it. It's not particularly easy without the pipeline automating secrets (and I don't think everyone will have all the things available/configured), but I'll add a blurb to the readme to help future contributors, too.

src/common.speech/DialogServiceAdapter.ts
trrwilson: That's not pedantic at all and I appreciate you calling it out! I was actually curious if string interpolation was possible and I should have looked more closely. Gladly adopted!

PR #302, opened 2021-01-12
add bookmark reached and viseme received event for TTS

yulin-li:


src/common.speech/SynthesisAdapterBase.ts
```diff
@@ -356,24 +359,59 @@ export class SynthesisAdapterBase implements IDisposable {
...
-                                    this.privSynthesisTurn.currentTextOffset);
-
-                                if (!!this.privSpeechSynthesizer.wordBoundary) {
-                                    try {
-                                        this.privSpeechSynthesizer.wordBoundary(this.privSpeechSynthesizer, ev);
-                                    } catch (error) {
-                                        // Not going to let errors in the event handler
-                                        // trip things up.
+                            switch (metadata.Type) {
+                                case MetadataType.WordBoundary:
```
glharper, line 363: I like this structure a lot better than testing against string literals!

src/common.speech/SynthesisAdapterBase.ts
```diff
@@ -356,24 +359,59 @@ export class SynthesisAdapterBase implements IDisposable {
...
+                                    this.privSynthesisTurn.onWordBoundaryEvent(metadata.Data.text.Text);
+
+                                    const ev: SpeechSynthesisWordBoundaryEventArgs = new SpeechSynthesisWordBoundaryEventArgs(
```
glharper, line 366: could "ev," "bev," and "vev" have more descriptive names, like "boundaryEventArgs," "bookmarkEventArgs," and "visemeEventArgs"?

src/common.speech/SynthesisAdapterBase.ts
```diff
@@ -356,24 +359,59 @@ export class SynthesisAdapterBase implements IDisposable {
...
+                                            // Not going to let errors in the event handler
+                                            // trip things up.
+                                        }
+                                    }
+                                    break;
+                                case MetadataType.Viseme:
+
+                                    this.privSynthesisTurn.onVisemeMetadataReceived(metadata);
+
+                                    if (metadata.Data.IsLastAnimation !== false) {
```
glharper, line 399: if (metadata.Data.IsLastAnimation) {

src/common.speech/SynthesisTurn.ts
```diff
@@ -68,6 +69,12 @@ export class SynthesisTurn {
         return this.privTextOffset;
     }
 
+    public get visemeAnimation(): string {
```
glharper, line 72: Should setters and getters have side effects, like clearing a private variable here?

src/common.speech/SynthesisTurn.ts
```diff
@@ -202,6 +211,12 @@ export class SynthesisTurn {
         this.updateTextOffset(text);
     }
 
+    public onVisemeMetadataReceived(metadata: ISynthesisMetadata): void {
+        if (metadata.Data.AnimationChunk !== undefined) {
+            this.privPartialVisemeAnimation = this.privPartialVisemeAnimation + metadata.Data.AnimationChunk;
```
glharper, line 216: Can use += here.

glharper (commented):
Can we update an existing test or add a new test to cover the new onAudioStart callback?

src/common.speech/SynthesisAdapterBase.ts
yulin-li: thanks for the detailed review!

src/common.speech/SynthesisAdapterBase.ts
yulin-li: thx, updated them

src/common.speech/SynthesisAdapterBase.ts
yulin-li: thx

src/common.speech/SynthesisTurn.ts
yulin-li: yes, it's hacky... I updated it to a method instead of a getter

yulin-li:
> Can we update an existing test or add a new test to cover the new onAudioStart callback?
Seems that the onAudioStart() would be triggered only in browser, and could not be tested in current test framework. I manually tested this method and it works.

glharper (approved):

PR #224, opened 2020-08-03
Update dependencies

orgads:

PR #310, opened 2021-01-26
update https-proxy-agent lib to version that doesn't depend on proble…

### LINKED ISSUE #309: Update https-proxy-agent
The version of https-proxy-agent (3.0.1) used in this package causes conflicts with other packages because it uses agent-base version that patches https global object (see https://github.com/TooTallNate/node-agent-base/issues/35). Please update package.json or merge #224 .

glharper:
…matic agent-base version. Closes #309

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2055,19 +2055,19 @@
       "dev": true
     },
     "agent-base": {
-      "version": "6.0.0",
-      "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-6.0.0.tgz",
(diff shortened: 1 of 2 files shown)


brandom-msft (approved):
I'll take your word for it :)

orgads:
Why not 5?

glharper:
@orgads IIRC there was an API change in 5 that didn't play well with our TS code.

orgads:
But I adapted it in this commit.

glharper:
@orgads By "this commit," are you referring to [this PR?]( https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/224/files) This repo has a policy of being very conservative with our dependency version updates, so as not to cause problems for customers using older versions of (Node/TypeScript/Library Foo).

orgads:
Yes, that's what I meant :)

Ok, I accept.

PR #311, opened 2021-01-27
ReplayableAudioNode: Fix exception on bad input

orgads:
If the offset value is wrong, and there not enough buffers,
an exception is thrown:
Cannot read property 'chunk' of undefined


orgads:
cc @bivas6 @glharper

glharper:
@orgads This looks good, and passes tests. Thanks for opening this.

glharper (approved):

PR #312, opened 2021-01-28
Websocket: Replace recursion with a loop

orgads:
sendItem holds a reference to the data on each level
in the recursion, and it is not released until close.

Reuse the same object instead.


orgads:
cc @bivas6 @yoshigev

glharper (approved):

glharper:
@orgads Thanks for submitting this. Again, this passes tests and seems reasonable.

glharper (approved):

PR #318, opened 2021-02-02
DialogServiceConnector: Restore support for Direct Line tokens

### LINKED ISSUE #315: BotFrameworkConfig.fromHost() should not set a SpeechServiceConnection_Host with a trailing slash
## Background

When calling `BotFrameworkConfig.fromHost('xyz')` or `BotFrameworkConfig.fromHost(new URL('wss://xyz.convai.speech.azure.us'))`, the Web Socket connection established contains double-slash and the server is returning 404.

In network trace, the URL it is calling become (note the double-slash after the hostname):

`wss://xyz.convai.speech.azure.us//api/v3?connectionid=...`

The correct one should be the following, without double-slash.

`wss://xyz.convai.speech.azure.us/api/v3?connectionid=...`

## Technical details

In [`/src/sdk/BotFrameworkConfig.ts:121`](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/master/src/sdk/BotFrameworkConfig.ts#L121), it set `SpeechServiceConnection_Host` with `new URL('wss://xyz.convai.speech.azure.us').toString()`. And it become `wss://xyz.convai.speech.azure.us/`. The trailing slash is added by `URL.toString()`.

When running under Node.js 14, you can also see same behavior:

image

Then, when [`src/common.speech/DialogConnectorFactory.ts:70`](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/master/src/common.speech/DialogConnectorFactory.ts#L70) construct the URL, it become double-slash.

When the code is run, the network trace I captured contains double-slash and the service is returning 404. Without the double-slash, the service operates normally.

image

image

## Workaround

After calling `BotFrameworkConfig.fromHost()`, manually set the `SpeechServiceConnection_Host` property.

### LINKED ISSUE #317: Direct Line Speech: "enableInternalHTTPSupport" flag is not working as service return 400 due to bad URL
In Direct Line Speech adapter, one of our test is to make sure the `enableInternalHTTPSupport` flag is working properly.

After we bumped to Speech SDK 1.15.0, with the flag enabled, we are seeing the Web Socket URL changed and the service always return 400 with the new URL.

Previously, the working URL is (broken into multiple lines for clarity):

```
https://westus2.convai.speech.microsoft.com/directline/api/v1?
  &Authorization=eyJhbGc...
  &format=simple
  &language=en-US
  &X-ConnectionId=3105F06...
  &X-DLS-Secret=ew0KICA...
```

The new URL that does not work is:

```
https://westus2.convai.speech.microsoft.com/directline/api/v1?
  &Authorization=eyJhbGc...
  &enableBotMessageStatus=true
  &format=simple
  &language=en-US
  &connectionid=613A427...
  &botid=ew0KICA...
```

Differences:

```diff
  https://westus2.convai.speech.microsoft.com/directline/api/v1?
    &Authorization=eyJhbGc...
+   &enableBotMessageStatus=true
    &format=simple
    &language=en-US
-   &X-ConnectionId=3105F06...
+   &connectionid=613A427...
-   &botid=ew0KICA...
+   &X-DLS-Secret=ew0KICA...
```

We tried to hack the code by reverting these 2 names and it worked properly:

image

It was caused by a recent PR changed the `QueryParameterNames.ts`, https://github.com/microsoft/cognitive-services-speech-sdk-js/commit/12c0e342f00050769b32c020459633ea6a342c38#diff-30a5c51a0d9988fc248b73ddac085bfdfcff208d0e471a767f2a8ed4d73753b6.

The code that use the `QueryParameterNames` lookup to generate the Web Socket URL can be found in [`DialogConnectorFactory.ts`](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/master/src/common.speech/DialogConnectorFactory.ts).

This issue seems not workaround-able and we are blocked on using Speech SDK 1.15.0.

trrwilson:
## Problem
1.15 fixed support for `botid` by appropriately mapping `Conversation_ApplicationId` to the `botid` query string parameter. This unwittingly broke an overloaded use of `Conversation_ApplicationId`: Direct Line token passthrough.

To do channel authentication, the agent context message currently grabs the `Conversation_ApplicationId` value.

* But providing a `botid` that isn't valid (not the Azure resource name) will cause DLS to issue a 4xx error (can't find the bot!)
* And we just made that map to said key

The correct solution here is to demux the concept of the extra token and the concept of a botid. Here that's done by adding a new property, `Conversation_Application_Connection_Id`. This will map into the agent config's `connectionId` field and appropriately pass through as the authentication token -- without messing with the ability to select non-default bots for a subscription.

Fixes #317

## Extras
* A related problem was found where using a strongly-typed URL as the host (needed for more flexible suffix support) caused an extraneous extra `/` to end up in the final connection URL. That's addressed by standardizing whether or not the forward slash is present in this fragment.

Fixes #315 

* The use of "connectionId" as a key, although not regressed anywhere, isn't right and the proper "X-ConnectionId" value is now used instead.

* Test coverage is also added to check that this property goes to the right place. We didn't have any events/monitoring for agent config messages (or anything in dialog, actually), so there's a bit of plumbing added.

### DIFF
--- a/tests/DialogServiceConnectorTests.ts
+++ b/tests/DialogServiceConnectorTests.ts
@@ -6,6 +6,7 @@ import {
     ConsoleLoggingListener,
     WebsocketMessageAdapter,
 } from "../src/common.browser/Exports";
+import { AgentConfig } from "../src/common.speech/Exports";
 import { HeaderNames } from "../src/common.speech/HeaderNames";
 import { QueryParameterNames } from "../src/common.speech/QueryParameterNames";
 import {
@@ -14,6 +15,7 @@ import {
     EventType,
     IDetachable,
     PlatformEvent,
+    SendingAgentContextMessageEvent,
 } from "../src/common/Exports";
 import { BotFrameworkConfig, OutputFormat, PropertyId, PullAudioOutputStream } from "../src/sdk/Exports";
 import { Settings } from "./Settings";
@@ -1112,6 +1114,46 @@ test("SendActivity fails with invalid JSON object", (done: jest.DoneCallback) =>
     });
 });
 
+describe('Agent config message tests', () => {
+    let eventListener: IDetachable;
+    let observedAgentConfig: AgentConfig;
+
+    beforeEach(() => {
+        eventListener = Events.instance.attachListener({
+            onEvent: (event: PlatformEvent) => {
+                if (event instanceof SendingAgentContextMessageEvent) {
+                    const agentContextEvent = event as SendingAgentContextMessageEvent;
+                    observedAgentConfig = agentContextEvent.agentConfig;
+                }
+            },
+        });
+    });
+
+    afterEach(async () => {
+        await eventListener.detach();
+        observedAgentConfig = undefined;
+    });
+
+    test('Agent connection id can be set', async (done: jest.DoneCallback) => {
+        const testConnectionId: string = "thisIsTheTestConnectionId";
+        const dialogConfig: sdk.BotFrameworkConfig = BuildBotFrameworkConfig();
+        dialogConfig.setProperty(sdk.PropertyId.Conversation_Agent_Connection_Id, testConnectionId);
+        objsToClose.push(dialogConfig);
+        const connector: sdk.DialogServiceConnector = BuildConnectorFromWaveFile(dialogConfig);
+        objsToClose.push(connector);
+
+        connector.listenOnceAsync(
+            () => {
+                expect(observedAgentConfig).not.toBeUndefined();
+                expect(observedAgentConfig.get().botInfo.connectionId === testConnectionId);
+                done();
+            },
+            () => {
+                done.fail();
+            });
+    });
+});
+
 describe.each([
     /* [
       <description>,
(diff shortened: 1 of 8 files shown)


BrianMouncer (approved):
:shipit:

tests/DialogServiceConnectorTests.ts
```diff
@@ -1112,6 +1114,46 @@ test("SendActivity fails with invalid JSON object", (done: jest.DoneCallback) =>
...
+        dialogConfig.setProperty(sdk.PropertyId.Conversation_Agent_Connection_Id, testConnectionId);
+        objsToClose.push(dialogConfig);
+        const connector: sdk.DialogServiceConnector = BuildConnectorFromWaveFile(dialogConfig);
+        objsToClose.push(connector);
+
+        connector.listenOnceAsync(
+            () => {
+                expect(observedAgentConfig).not.toBeUndefined();
+                expect(observedAgentConfig.get().botInfo.connectionId === testConnectionId);
+                done();
```
rhurey, line 1149: Because this is in a callback, you want a try / catch / fail pattern here. Else an error will just show up as a timeout.

rhurey (approved):
:shipit:

glharper (approved):

brandom-msft (approved):

Issue #264, opened 2020-10-30
Expo bundle fails with dependancy resolution error


anilsamuel:
Get the error below in expo; shall I open a new issue ?

Failed building JavaScript bundle.
The package at "node_modules/microsoft-cognitiveservices-speech-sdk/distrib/lib/src/common.browser/CertChecks.js" attempted to import the Node standard library module "url". It failed because the native React runtime does not include the Node standard library. Read more at https://docs.expo.io/workflow/using-libraries/#using-third-party-libraries

package.json below if that helps

```
{
  "dependencies": {
    "@expo-google-fonts/inter": "^0.1.0",
    "@expo/vector-icons": "^10.0.0",
    "@react-native-community/masked-view": "0.1.10",
    "@react-native-community/netinfo": "5.9.6",
    "@react-native-community/picker": "1.6.6",
    "@react-navigation/native": "^5.8.2",
    "expo": "^39.0.3",
    "expo-barcode-scanner": "~9.0.0",
    "expo-constants": "~9.2.0",
    "expo-image-picker": "~9.1.0",
    "expo-status-bar": "~1.0.2",
    "microsoft-cognitiveservices-speech-sdk": "^1.14.0",
    "react": "~16.13.0",
    "react-dom": "~16.13.0",
    "react-native": "https://github.com/expo/react-native/archive/sdk-39.0.2.tar.gz",
    "react-native-camera": "^3.40.0",
    "react-native-elements": "^3.0.0-alpha.1",
    "react-native-gesture-handler": "~1.7.0",
    "react-native-qrcode-svg": "^6.0.6",
    "react-native-reanimated": "~1.13.0",
    "react-native-safe-area-context": "3.1.4",
    "react-native-screens": "~2.10.1",
    "react-native-size-matters": "^0.3.1",
    "react-native-svg": "12.1.0",
    "react-native-web": "^0.14.7",
    "react-navigation": "^4.4.3",
    "react-navigation-stack": "^2.10.0"
  },
  "devDependencies": {
    "@babel/core": "~7.9.0"
  }
}
```

anilsamuel:
url, util and https packages were compalined about and was installed using yarn add for each of them and eventually bricked as below

```
Logs for your project will appear below. Press Ctrl+C to exit.
Installing Expo client on device
Opening exp://192.168.20.61:19000 on Pixel_3a_API_30_x86
Failed building JavaScript bundle.
[...]

PR #322, opened 2021-02-04
replace url dependency with url-parse from CertChecks.ts

glharper:
The node dependency on url has tripped up a couple of customers (for example [see here](https://github.com/microsoft/cognitive-services-speech-sdk-js/issues/264) ), and the library seems a little less than ideal as a dependency (for example, [see this file](https://github.com/defunctzombie/node-url/blob/master/url.js) ), so I converted the functionality needed in CertsCheck to TypeScript and cleaned up a bit of the code.


trrwilson:
Is this the way other projects handle it? I'm always leery about duplicating this much "boilerplate" logic since it never ends up being as simple as you'd hope. I know JS can be infuriating at times and this may be such an instance, but it's always good to check: is there no other way to get "URL stuff" in without writing our own URL parser?

One hack I've seen used is to programmatically create an 'a' element in a DOM and then do the parsing from that. Do we have any access to that as an alternative? (Just brainstorming -- far from a JS guru here!)

glharper:
> Is this the way other projects handle it? I'm always leery about duplicating this much "boilerplate" logic since it never ends up being as simple as you'd hope. I know JS can be infuriating at times and this may be such an instance, but it's always good to check: is there no other way to get "URL stuff" in without writing our own URL parser?
> 
> One hack I've seen used is to programmatically create an 'a' element in a DOM and then do the parsing from that. Do we have any access to that as an alternative? (Just brainstorming -- far from a JS guru here!)

I've looked at the DOM url parser and url-parse lib in turn, and both do not parse urls with the same interfaces as the node url lib, and unfortunately neither passes our OSCP tests. Am absolutely open to other libs, just couldn't find anything matching the functionality of node-url.

yulin-li (approved):

yulin-li:
nit: better to update the PR title as you are replacing a dependency

Issue #320, opened 2021-02-03
iOS Safari: STT Error.


BryanDollery:
I've written a react spa pwa targetting web, android, and iOS. My app works without a problem on the first two platforms, but it doesn't work on a fully updated iPhone 11. The error message is:

```
Unhandled Rejection (TypeError): null is not an object (evaluating 'this.privInitializeDeferral.reject')
(anonymous function)
src/common.browser/MicAudioSource.ts:141
  138 |             this.privInitializeDeferral.resolve();
  139 |         }, (error: MediaStreamError) => {
  140 |             const errorMsg = `Error occurred during microphone initialization: ${error}`;
> 141 |             this.privInitializeDeferral.reject(errorMsg);
      | ^
  142 |             this.onEvent(new AudioSourceErrorEvent(this.privId, errorMsg));
  143 |         });
  144 | }
```
This is a short commercial project for a major European MS customer, we're a partner, and delivery is due in a few days. I could take this through formal channels but I thought that it might be quicker to report the issue here. I'm running v1.15.0 of the sdk.

I have written this up in more detail on StackOverflow:

https://stackoverflow.com/questions/66028639/microsoft-cognitiveservices-speech-sdk-javascript-on-ios-safari-micaudiosource

TIA

glharper:
@BryanDollery We don't officially support React integrations like this, but can you reproduce this using version 1.14.1? This may be a regression in 1.15

BryanDollery:
Hi @glharper, thanks for the response. This problem isn't specific to the nature of the app or React. It's a javascript/safari problem. The react stuff is only about how to structure the javascript and UI. The problem seems to be in the cog-services SDK for javascript. I think it's not waiting for permission to use the microphone before trying to actually use it. One big issue here is that I can't debug a pwa on iOS safari to get a full stack-trace.

glharper:
@BryanDollery. Understood, but there was a change to the error handling in that specific section of code in MicAudioSource.ts for 1.15. If you could try to repro with 1.14.1, that would at least tell me whether that change is to blame (and possibly give you a workaround until we release a fix).

BryanDollery:
Ok, tried it with 1.14.1, and the error has changed but it's still basically the same underlying error:

```
Unhandled Rejection (TypeError): null is not an object (evaluating 'tmp.reject')
(anonymous function)
src/common.browser/MicAudioSource.ts:147
  144 | // without a lot of code replication.
[...]

PR #323, opened 2021-02-05
Correctly handle initialization promise when browser forces mic off b…

glharper:
…efore turnOn() completes #320


ralph-msft (approved):
:shipit:

PR #327, opened 2021-02-16
Correct microphone timing issue if mic stream ends occurs before stop…

glharper:
…Recognizer.

From https://portal.microsofticm.com/imp/v3/incidents/details/228051095/home.

Essentially the new FireFox is getting to the end of the mic stream (when close() is called on it) before stopRecognizing has been called. This accounts for that possibility.


brandom-msft (approved):
👍

Issue #314, opened 2021-01-30
Property "SpeechServiceConnection_SynthOutputFormat" is ignored in DialogServiceAdapter


compulim:
The code handling `SpeechServiceConnection_SynthOutputFormat` is wrong and the cause starts at [`DialogServiceAdapter.ts:599`](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/master/src/common.speech/DialogServiceAdapter.ts#L599).

```js
const pullAudioOutputStream: PullAudioOutputStreamImpl = turn.processActivityPayload(
  activityPayload,
  (SpeechSynthesisOutputFormat as any)[this.privDialogServiceConnector.properties.getProperty(PropertyId.SpeechServiceConnection_SynthOutputFormat, undefined)]
);
```

Then, it call the [`DialogServiceTurnState.processActivityPayload` function](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/master/src/common.speech/DialogServiceTurnState.ts#L32), it accept `AudioOutputFormatImpl` instead of `SpeechSynthesisOutputFormat`.

```js
public processActivityPayload(payload: ActivityPayloadResponse, audioFormat?: AudioOutputFormatImpl): PullAudioOutputStreamImpl {
    if (payload.messageDataStreamType === MessageDataStreamType.TextToSpeechAudio) {
        this.privAudioStream = AudioOutputStream.createPullStream() as PullAudioOutputStreamImpl;
        this.privAudioStream.format = (audioFormat !== undefined) ? audioFormat : AudioOutputFormatImpl.getDefaultOutputFormat();
        // tslint:disable-next-line:no-console
        // console.info("Audio start debugturn:" + this.privRequestId);
    }
    return this.privAudioStream;
}
```

Then it call the [setter of `PullAudioOutputStream.format`](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/master/src/sdk/Audio/AudioOutputStream.ts#L120):

```js
public set format(format: AudioStreamFormat) {
    if (format === undefined || format === null) {
        this.privFormat = AudioOutputFormatImpl.getDefaultOutputFormat();
    }
    this.privFormat = format as AudioOutputFormatImpl;
}
```

[...]

PR #329, opened 2021-02-17
Fix audio playback for iOS Safari

brandom-msft:
iOS is heavily (related to other browser functionality) restricted in media playback capabilities, especially when using webkitAudio. For example, decodeAudioData is restrictive and selective on the media it supports (while also being sparse in documentation about said functionality).

I spent some time looking at how we could provide the audio data in a way that BaseAudioContext would be happy with it, but in the end found we can actually stop using it altogether and instead directly provide the audio data as we do in other browsers.

I've also added a commit that addresses the audio format issue #314

### DIFF
--- a/src/sdk/Audio/BaseAudioPlayer.ts
+++ b/src/sdk/Audio/BaseAudioPlayer.ts
@@ -39,16 +39,12 @@ export class BaseAudioPlayer {
      */
     public playAudioSample(newAudioData: ArrayBuffer, cb?: () => void, err?: (error: string) => void): void {
         marshalPromiseToCallbacks((async (): Promise<void> => {
-            if (!!(window as any).webkitAudioContext) {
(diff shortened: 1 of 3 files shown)


src/sdk/Audio/BaseAudioPlayer.ts
```diff
@@ -39,16 +39,12 @@ export class BaseAudioPlayer {
      */
     public playAudioSample(newAudioData: ArrayBuffer, cb?: () => void, err?: (error: string) => void): void {
         marshalPromiseToCallbacks((async (): Promise<void> => {
-            if (!!(window as any).webkitAudioContext) {
-                await this.playAudio(newAudioData);
```
brandom-msft, line 43: I'll remove the playAudio function altogether (found below ~line 159)

glharper (approved):

PR #348, opened 2021-03-23
Glharper/voice profile ops 338

glharper:
Voice Profile - Add API for REST operations requested in #338 and updates to use latest Speaker Recognition API ~~and endpoints~~


src/sdk/SpeakerRecognitionResult.ts
```diff
@@ -35,12 +35,12 @@ export class SpeakerRecognitionResult {
                 const json: { identifiedProfile: { profileId: string, score: number } } = JSON.parse(data);
                 Contracts.throwIfNullOrUndefined(json, "JSON");
                 this.privProfileId = json.identifiedProfile.profileId;
-                this.privScore = json.identifiedProfile.score;
+                // this.privScore = json.identifiedProfile.score;
```
brandom-msft, line 38: looks like this was replaced by privConfidence, safe to delete?

src/common.speech/SpeakerIdMessageAdapter.ts
```diff
@@ -70,14 +69,17 @@ export class SpeakerIdMessageAdapter {
      * @public
      * @returns {Promise<IRestResponse>} rest response to enrollment request.
      */
-    public createEnrollment(profile: VoiceProfile, audioSource: IAudioSource):
-        Promise<IRestResponse> {
+    public async createEnrollment(profile: VoiceProfile, audioSource: IAudioSource): Promise<IRestResponse> {
```
brandom-msft, line 72: should the function be named with an "Async" suffix?

src/sdk/VoiceProfileClient.ts
```diff
@@ -80,6 +81,62 @@ export class VoiceProfileClient {
...
+    /**
+     * Get current information of a voice profile
+     * @member VoiceProfileClient.prototype.retrieveEnrollmentResultAsync
+     * @function
+     * @public
+     * @param {VoiceProfile} profile Voice Profile to retrieve info for
+     * @param cb - Callback invoked once Voice Profile has been created.
+     * @param err - Callback invoked in case of an error.
+     */
+    public retrieveEnrollmentResultAsync(profile: VoiceProfile, cb?: (e: VoiceProfileEnrollmentResult) => void, err?: (e: string) => void): void {
```
brandom-msft, line 93: similar question to above, but the opposite.. When do we signal "Async" on function names? Is it indicated by the signature or by the calls occurring inside it?

src/sdk/VoiceProfileEnrollmentResult.ts
```diff
@@ -28,48 +28,111 @@ export interface IEnrollmentResultDetails {
...
     }
 
     public get errorDetails(): string {
         return this.privErrorDetails;
     }
+
+    private static getIdentificationDetails(json: any, profileId?: string): any {
+        return {
+            audioSpeechLength: json.speechTime ? parseFloat(json.speechTime) : null,
+            enrollmentLength: parseFloat(json.enrollmentSpeechTime),
```
brandom-msft, line 121: enrollmentLength and remainingEnrollmentSpeechLength - we should guard against NaN on those as well? Or do we necessarily need to? In any case, I'm wondering if all or none should, for consistency

brandom-msft (approved):
Looks good, most of of my comments are general inquiry so not gating check-in approval on them. (But I am curious to know the answers :) )

src/sdk/VoiceProfileClient.ts
glharper: In this case, parity with Native API, which has a RetrieveEnrollmentResultAsync method on VoiceProfileClient. Unfortunately our "...Async" named APIs typically are not JS async, as they only offer parameterized callbacks. Currently the plan is to make the breaking change to async methods for v2.0.

PR #356, opened 2021-04-07
Add basic Keyword Verification support to DialogServiceConnector

trrwilson:
### Description
The Speech Service for `DialogServiceConnector` supports service-side *Keyword Verification* (KWV). This confirms that audio you provide contains the keyword you specify and gives an explicit acceptance or rejection depending on the result.

API design for exposing this in a more friendly way is forthcoming, but this change introduces a means to use KWV via properties that's at parity with the mechanism available on the C++/C#/Java/etc. SDK.

### Usage details

KWV is used on the `DialogServiceConnector` by including a few properties on its `BotFrameworkConfig` or `CustomCommandsConfig`, as set by `config.setProperty(name, value)`:

| Property | Description | Example |
|---|---|---|
| `SPEECH-KeywordsToDetect` | A semicolon-delimited list of keywords to test the input audio against. If any of these keywords are detected at the start of the audio within the allotted time, the utterance will be accepted; otherwise, it'll be rejected. | `"contoso"` OR `"contoso;hey banana"` |
| `SPEECH-KeywordsToDetect-Durations` | A semicolon-delimited list of durations, *as specified in hundred-nanosecond (HNS) units*, that correspond to keywords provided in `SPEECH-KeywordsToDetect` and constrain the duration of audio that will be evaluated. If this duration is reached without a positive evaluation having occurred, the utterance will be rejected even if it hasn't arrived at an explicit rejection yet. | `"20000000"` (2s) |
| `SPEECH-KeywordsToDetect-Offsets` | **Not yet supported.** May later specify the length of an initial segment of audio to "skip" evaluation for, on a per-keyword basis. | |

When a keyword is verified by the speech service, a `recognized` event *without a promise resolution* will be raised and provide a `ResultReason` of `RecognizedKeyword`. Evaluation of the utterance will then proceed as usual, with a final result (and promise resolution) assigned a reason of `RecognizedSpeech` finalizing the turn.

When a keyword is rejected by the speech service, a `recognized event` *with a promise resolution* will be raised a provide a `ResultReason` of `NoMatch`. No further evaluation of audio will be done after this point and the turn is over.


rhurey:
No tests?

rhurey:
Ah, missed it was a draft.

---
In reply to: [814560486](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/356#issuecomment-814560486) [](ancestors = 814560486)

glharper (approved):

PR #359, opened 2021-04-16
update speech synthesis format

yulin-li:
- Add `Raw24Khz16BitMonoTrueSilk`, mark siren format as unsupported
- Call `onAudioEnd()` callback if the format is not supported for playback
- Update riff header for mulaw

### DIFF
--- a/src/sdk/Audio/SpeakerAudioDestination.ts
+++ b/src/sdk/Audio/SpeakerAudioDestination.ts
@@ -10,13 +10,15 @@ import {
     INumberDictionary
 } from "../../common/Exports";
 import { AudioStreamFormat, IPlayer } from "../Exports";
-import { AudioFormatTag, AudioOutputFormatImpl } from "./AudioOutputFormat";
+import { AudioOutputFormatImpl } from "./AudioOutputFormat";
 import { PullAudioOutputStreamImpl } from "./AudioOutputStream";
+import { AudioFormatTag } from "./AudioStreamFormat";
 
 const MediaDurationPlaceholderSeconds = 60 * 30;
 
 const AudioFormatToMimeType: INumberDictionary<string> = {
     [AudioFormatTag.PCM]: "audio/wav",
+    [AudioFormatTag.MuLaw]: "audio/x-wav",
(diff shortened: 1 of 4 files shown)


src/sdk/Audio/SpeakerAudioDestination.ts
```diff
@@ -85,27 +87,40 @@ export class SpeakerAudioDestination implements IAudioDestination, IPlayer {
...
-                const audioBlob = new Blob([receivedAudio], { type: AudioFormatToMimeType[this.privFormat.formatTag] });
-                this.privAudio.src = window.URL.createObjectURL(audioBlob);
-                this.notifyPlayback().then(() => {
-                    if (!!cb) {
-                        cb();
+            } else {
+                let receivedAudio = new ArrayBuffer(this.privBytesReceived);
+                this.privAudioOutputStream.read(receivedAudio).then((_: number): void => {
+                    if (this.privFormat.hasHeader) {
+                        receivedAudio = SynthesisAdapterBase.addHeader(receivedAudio, this.privFormat);
```
glharper, line 100: Doesn't addHeader already perform the hasHeader() check?

yulin-li: The `receivedAudio` is from `privAudioOutputStream` and has no header. So if the format needs a header, we need to add it here and send the audio chunk with header to browser for playback.

glharper: My comment is about the line:
 `
if (this.privFormat.hasHeader) {`

The first three lines of SynthesisAdapterBase.addHeader are:
```
    public static addHeader(audio: ArrayBuffer, format: AudioOutputFormatImpl): ArrayBuffer {
        if (!format.hasHeader) {
            return audio;
        }
```
Do we need to perform the hasHeader check twice?

yulin-li: Yes, you are right, I removed the duplicate checking. Thanks!

glharper (approved):

PR #365, opened 2021-04-22
Fix UnhandledPromiseRejection warning in Node

glharper:
For long connection scenarios, ServiceAdapterBase can receive a turn.start message when the current turn hasn't finished, causing it to reject a promise that's not being awaited, resulting in a UnhandledPromiseRejection warning in Node. This PR handles the rejection to avoid that Unhandled warning (with no adverse effects if the promise is being awaited).

### DIFF
--- a/src/common.speech/RequestSession.ts
+++ b/src/common.speech/RequestSession.ts
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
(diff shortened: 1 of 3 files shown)


glharper:
(Note: technically it's the reassignment of this.privTurnDeferral that causes the Unhandled warning, since the rejection of the previous promise isn't handled before the reassignment.)

brandom-msft (approved):

PR #375, opened 2021-05-18
fix for websocket msg size overage

glharper:
Fix https://github.com/Azure-Samples/cognitive-services-speech-sdk/issues/1079


rhurey (approved):
:shipit:

PR #381, opened 2021-06-02
implement getActivationPhrasesAsync with test

glharper:
Add getActivationPhrasesAsync API to support v3 of Speaker Recognition backend

### DIFF
--- a/src/sdk/VoiceProfilePhraseResult.ts
+++ b/src/sdk/VoiceProfilePhraseResult.ts
@@ -9,16 +9,16 @@ import {
 
 /**
  * Output format
- * @class VoiceProfileAuthorizationPhraseResult
+ * @class VoiceProfilePhraseResult
  */
-export class VoiceProfileAuthorizationPhraseResult extends VoiceProfileResult {
+export class VoiceProfilePhraseResult extends VoiceProfileResult {
     private privPhrases: string[] = [];
 
     public constructor(reason: ResultReason, statusText: string, json: any) {
         super(reason, statusText);
         Contracts.throwIfNullOrUndefined(json, "phrases array");
         for (const item of json) {
-            this.privPhrases.push(item.passPhrase);
+            this.privPhrases.push(item.passPhrase || item.activationPhrase);
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/sdk/VoiceProfilePhraseResult.ts
```diff
@@ -9,16 +9,16 @@ import {
...
-export class VoiceProfileAuthorizationPhraseResult extends VoiceProfileResult {
+export class VoiceProfilePhraseResult extends VoiceProfileResult {
     private privPhrases: string[] = [];
 
     public constructor(reason: ResultReason, statusText: string, json: any) {
         super(reason, statusText);
         Contracts.throwIfNullOrUndefined(json, "phrases array");
         for (const item of json) {
-            this.privPhrases.push(item.passPhrase);
+            this.privPhrases.push(item.passPhrase || item.activationPhrase);
```
brandom-msft, line 21: TIL

brandom-msft (approved):

src/sdk/VoiceProfilePhraseResult.ts
brandom-msft: Does the documentation detail anything about the use of one or the other? What I'm thinking about is the expectation if both are provided.. I assume the OR will shortcut on the first one and ignore the second - would this matter?

glharper: as it stands now, passPhrases will only exist for text dependent verification, and activationPhrases for the other two SPID types. The schema properties in question are listed [here ](https://dev.azure.com/msasg/Skyman/_git/speaker-recognition-frontend?path=%2Fmanagement%2FAPI%2FV3%2Fverification-swagger.yaml&version=GBmaster&line=621&lineEnd=622&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) and [here](https://dev.azure.com/msasg/Skyman/_git/speaker-recognition-frontend?path=%2Fmanagement%2FAPI%2FV3%2Fverification-swagger.yaml&version=GBmaster&line=994&lineEnd=995&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

PR #388, opened 2021-06-24
Don't stop tracks on input stream, just disconnect nodes. 

### LINKED ISSUE #385: Once I call stopContinuousRecognitionAsync, the mediarecorder sharing the same stream will not work properly
hello, after I clone the mediarecorder stream and pass it to TranslationRecognizer, once I use stopContinuousRecognitionAsync, the mediarecorder cannot record normally. Can you tell me why?
`function initRecorder() {
    navigator.mediaDevices
      .getUserMedia(mediaOptions)
      .then((stream) => {
        if (type === RecorderType.ALL) {
          videoRef.current.srcObject = stream;
          recorder.current = new MediaRecorder(stream);

          recorder.current.ondataavailable = function (e) {
            if (e.data && e.data.size > 0) {
              console.log("MediaRecorder.ondataavailable");
              chunks.push(e.data);
            }
          };
          recorder.current.onstop = (e) => {
            let blob = new Blob(chunks, { type: "video/mp4" });
            chunks = [];
            if (typeof props.onRecorderFinished === "function") {
              props.onRecorderFinished(blob);
            }
          };

          let audioStream = stream.clone();
          let mediaAudioStream = new MediaStream(audioStream.getAudioTracks());
          initRecognizer(mediaAudioStream);
  }`

glharper:
Fix #385.

### DIFF
--- a/src/common.browser/PCMRecorder.ts
+++ b/src/common.browser/PCMRecorder.ts
@@ -107,7 +107,6 @@ export class PcmRecorder implements IRecorder {
             }
             if (this.privMediaResources.source) {
                 this.privMediaResources.source.disconnect();
-                this.privMediaResources.stream.getTracks().forEach((track: any) => track.stop());
                 this.privMediaResources.source = null;
             }
         }
--- a/src/common.browser/PCMRecorder.ts
+++ b/src/common.browser/PCMRecorder.ts
@@ -7,6 +7,11 @@ import { IRecorder } from "./IRecorder";
 export class PcmRecorder implements IRecorder {
     private privMediaResources: IMediaResources;
     private privSpeechProcessorScript: string; // speech-processor.js Url
+    private privStopInputOnRelease: boolean;
+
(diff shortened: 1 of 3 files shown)


src/common.browser/PCMRecorder.ts
```diff
@@ -107,7 +107,6 @@ export class PcmRecorder implements IRecorder {
             }
             if (this.privMediaResources.source) {
                 this.privMediaResources.source.disconnect();
-                this.privMediaResources.stream.getTracks().forEach((track: any) => track.stop());
```
rhurey, line 110: >his.privMediaResources.stream.getTracks().forEach((track: any) => track.stop()); [](http://example.com/codeflow?start=17&length=80)

Tried this in a browser, and the problem is now when the microphone is used directly by the SDK, the microphone light stays after recognizing speech.

glharper: Corrected, please try now.

rhurey (approved):
:shipit:

PR #392, opened 2021-07-01
Glharper/conversation translation fixes 389

### LINKED ISSUE #389: Multi-device conversation: Leaving participant microphone open results in disconnect after a few minutes
Using the sample [here](https://github.com/Azure-Samples/cognitive-services-speech-sdk/blob/master/quickstart/javascript/browser/multi-device-conversation/multi-device-conversation.html),
- open sample page in three separate tabs 
- host conversation in first tab, join as participant in other two
- click start speaking in all tabs and allow microphone access

After a few minutes without talking, at least one participant will disconnect. (I'd expect all participants to disconnect about the same time, as well as the host, if there's a set silence timeout.)

More issues
- When all participants disconnect, there's no way to rejoin the conversation using the sample without reloading the page.
- Even after reloading, trying to rejoin results in a 400 error: participant name already exists and is in the conversation.
- The host eventually gets a 404: resource does not exist, and is dropped.

glharper:
Fix #389 - correctly handle auth token change, send keep_alive message to conversation WS every 60 seconds

### DIFF
--- a/src/sdk/Transcription/Conversation.ts
+++ b/src/sdk/Transcription/Conversation.ts
@@ -321,22 +321,8 @@ export class ConversationImpl extends Conversation implements IDisposable {
             Contracts.throwIfNullOrUndefined(this.privRoom, this.privErrors.permissionDeniedConnect);
             // connect to the conversation websocket
             this.privParticipants.meId = this.privRoom.participantId;
-            this.privConversationRecognizer = ConversationRecognizerFactory.fromConfig(this.privConfig);
+            this.privConversationRecognizer = ConversationRecognizerFactory.fromConfig(this, this.privConfig);
 
-            // Because ConversationTranslator manually sets up and manages the connection, Conversation
-            // has to forward serviceRecognizer connection events that usually get passed automatically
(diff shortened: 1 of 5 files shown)


src/sdk/Transcription/Conversation.ts
```diff
@@ -895,7 +863,7 @@ export class ConversationImpl extends Conversation implements IDisposable {
         }
     }
 
-    private onTranslationReceived = (r: ConversationRecognizer, e: ConversationReceivedTranslationEventArgs): void => {
+    public onTranslationReceived = (r: ConversationRecognizer, e: ConversationReceivedTranslationEventArgs): void => {
```
rhurey, line 866: >onTranslationReceived [](http://example.com/codeflow?start=11&length=21)

Why are these now public on the SDK surface?

glharper: They're part of ConversationImpl, so they're only available after recasting, much like the recognizerImpl methods.

yulin-li (approved):

PR #397, opened 2021-07-07
Use AudioWorkletNode by default for mic input

### LINKED ISSUE #391: Replace PCM Recorder script processor with audio worklet
We are getting deprecation warnings in the console due to the use of script processor in pcm recorder. We've also noticed that on lower end devices (or throttling the CPU in chrome devtools) the quality of transcriptions suffers could using an audio worklet help?

glharper:
Fix #391

### DIFF
--- a/src/audioworklet/speech-processor.js
+++ b/src/audioworklet/speech-processor.js
@@ -10,7 +10,7 @@ class SpeechProcessor extends AudioWorkletProcessor {
 
   process(inputs, outputs) {
     const input = inputs[0];
-    const output = outputs[0];
+    const output = []; // Make sure output array can be assigned to
     for (let channel = 0; channel < input.length; channel += 1) {
       output[channel] = input[channel];
     }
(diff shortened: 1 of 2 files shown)


glharper:
Note there are differences between webWorkers and audioWorkletNodes. For example, webWorkers allow generated URL access on non-http URLs without generating a CORS access error, but audioWorkletNode does not. This PR keeps ScriptProcessorNode as a fallback when a CORS exception is thrown (this will be mostly be when a browser sample is opened from the filesystem, and not on an http server).

glharper:
https://glharper-js-test.azurewebsites.net/all.html to test mic input

rhurey (approved):
:shipit:

rssfrncs:
When can we expect this in a release? Thanks 🙏🏻

glharper:
@rssfrncs Our next release, 1.18, is expected within the next two weeks. This should be included.

PR #402, opened 2021-07-13
Glharper/audio input reattach fix

glharper:
FF threw error when referencing srcFrame.length if srcFrame is null/undefined.


BrianMouncer (approved):
look good.  only pet peeve is having multiple returns from a functions, but that's not entirely your fault... :-)

PR #404, opened 2021-07-15
Glharper/translation update target languages

glharper:


src/sdk/TranslationRecognizer.ts
```diff
@@ -268,4 +314,14 @@ export class TranslationRecognizer extends Recognizer {
         return new TranslationServiceRecognizer(authentication, connectionFactory, configImpl, recognizerConfig, this);
     }
 
+    private updateLanguages(languages: string[]): void {
+        const conn: Connection = Connection.fromRecognizer(this);
+        if (!!conn) {
+            const langs: string[] = languages.map((l: string) => l.substring(0, 2));
```
ralph-msft: >languages.map((l: string) => l.substring(0, 2)); [](http://example.com/codeflow?start=36&length=48)

I would recommend passing through the language codes the user provides as is instead of trying to parse them

ralph-msft (approved):
:shipit:

PR #410, opened 2021-07-30
Fix for regression, and test

glharper:
Fix [issue with detailed continuous recognition not stopping correctly](https://github.com/Azure-Samples/cognitive-services-speech-sdk/issues/1190)


rhurey (approved):
:shipit:

PR #412, opened 2021-08-04
Fix issue where audio context is closed twice on FF

glharper:
Either the FF JS engine is now super fast, or the audioContext close takes longer to return than for Chromium. This fixes a race condition in MicAudioSource where:
- await this.privContext.close() is called, takes a while to return, so in the meantime
- stream.detach() is called, which calls this.turnOff(), which calls this.destroyAudioContext, which calls this.privContext.close() again


BrianMouncer (approved):
This looks good to me, so long as privIsClosing is thread safe.

PR #419, opened 2021-08-16
Translation: Authentication refresh token not updating

ej-msl:
Websocket headers using original token as when config was initialized.

PR #420, opened 2021-08-16
add token refresh test for translation recognizer

glharper:
Add test for issue found in #419, translation recognizer auth token refresh not working correctly

### DIFF
--- a/tests/LongRunning/SpeechRecoAuthTokenRefreshTests.ts
+++ b/tests/LongRunning/SpeechRecoAuthTokenRefreshTests.ts
@@ -204,3 +204,175 @@ test("AuthToken refresh works correctly", (done: jest.DoneCallback) => {
             });
     });
 }, 1000 * 60 * 25); // 25 minutes.
+
+test("AuthToken refresh works correctly for Translation Recognizer", (done: jest.DoneCallback) => {
+    // tslint:disable-next-line:no-console
+    console.info("Name: AuthToken refresh works correctly for Translation Recognizer");
+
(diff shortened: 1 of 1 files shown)


tests/LongRunning/SpeechRecoAuthTokenRefreshTests.ts
```diff
@@ -204,3 +204,175 @@ test("AuthToken refresh works correctly", (done: jest.DoneCallback) => {
...
+                        if (bytesSent >= 32000) {
+                            bytesSent = 0;
+                            pumpSilence = false;
+                        }
+                        return buffer.byteLength;
+                    } else {
+                        const copyArray: Uint8Array = new Uint8Array(buffer);
+                        const start: number = bytesSent;
+                        const end: number = buffer.byteLength > (fileBuffer.byteLength - bytesSent) ? (fileBuffer.byteLength - 1) : (bytesSent + buffer.byteLength - 1);
+                        copyArray.set(new Uint8Array(fileBuffer.slice(start, end)));
```
brandom-msft, line 264: is it necessary to "new" an array, here? Won't slice hand that off?

brandom-msft (approved):
thanks for the nice use of comments throughout, they were helpful in following along

tests/LongRunning/SpeechRecoAuthTokenRefreshTests.ts
glharper: slice() gives an ArrayBuffer, which confusingly is not ArrayLike. (set() expects an ArrayLike parameter).

PR #422, opened 2021-08-23
Glharper/voice profile async await

glharper:
This is a breaking change to the Voice Profile operations, currently in preview/beta, making all public APIs in VoiceProfileClient and SpeakerRecognizer async/await (instead of taking function callback parameters).


brandom-msft (approved):

tests/VoiceProfileClientTests.ts
```diff
@@ -91,16 +91,14 @@ test("VoiceProfileClient with Bad credentials throws meaningful error", (done: j
...
-            done();
-        });
+    try {
+        const res: sdk.VoiceProfile = await r.createProfileAsync(type, "en-us");
+    } catch (error) {
+        const expectedCode: number = 401;
+        const expectedMessage: string = "Access denied due to invalid subscription key or wrong API endpoint. Make sure to provide a valid key for an active subscription and use a correct regional API endpoint for your resource.";
+        expect(error.toString()).toEqual(`Error: createProfileAsync failed with code: ${expectedCode}, message: ${expectedMessage}`);
+        done();
+    }
```
BrianMouncer, line 101: what if it does not throw at all for some reason?
Do you wnat to detect and fail for that?

BrianMouncer (approved):
:shipit:

PR #423, opened 2021-08-23
add "bearer " to front of token

glharper:
I will be adding a test to this as well

### DIFF
--- a/src/common.speech/CognitiveTokenAuthentication.ts
+++ b/src/common.speech/CognitiveTokenAuthentication.ts
@@ -24,10 +24,10 @@ export class CognitiveTokenAuthentication implements IAuthentication {
     }
 
     public fetch = (authFetchEventId: string): Promise<AuthInfo> => {
-        return this.privFetchCallback(authFetchEventId).then((token: string) => new AuthInfo(AuthHeader, token));
+        return this.privFetchCallback(authFetchEventId).then((token: string) => new AuthInfo(AuthHeader, "bearer " + token));
     }
 
     public fetchOnExpiry = (authFetchEventId: string): Promise<AuthInfo> => {
-        return this.privFetchOnExpiryCallback(authFetchEventId).then((token: string) => new AuthInfo(AuthHeader, token));
+        return this.privFetchOnExpiryCallback(authFetchEventId).then((token: string) => new AuthInfo(AuthHeader, "bearer " + token));
     }
 }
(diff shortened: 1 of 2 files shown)


brandom-msft (approved):

src/common.speech/CognitiveTokenAuthentication.ts
```diff
@@ -24,10 +24,10 @@ export class CognitiveTokenAuthentication implements IAuthentication {
     }
 
     public fetch = (authFetchEventId: string): Promise<AuthInfo> => {
-        return this.privFetchCallback(authFetchEventId).then((token: string) => new AuthInfo(AuthHeader, token));
+        return this.privFetchCallback(authFetchEventId).then((token: string) => new AuthInfo(AuthHeader, "bearer " + token));
```
BrianMouncer, line 27: >"bearer " [](http://example.com/codeflow?start=105&length=9)

Is it common practice to duplicate hard coded strings like this?

in our other API surfaces, these would all be pulled out the a webheaders and cookies .h file and defined as a reusable constant.

glharper: I think you're correct, this should be a class level constant. Will update the other PR to update this.

tests/SpeechRecognizerTests.ts
```diff
@@ -1756,6 +1756,37 @@ describe.each([true])("Service based tests", (forceNodeWebSocket: boolean) => {
...
+            console.info("Name: Endpoint URL With Auth Token Bearer added");
+
+            const fakeToken: string = createNoDashGuid();
+            const s: sdk.SpeechConfig = sdk.SpeechConfig.fromAuthorizationToken(fakeToken, "westus");
+            objsToClose.push(s);
+
+            const r: sdk.SpeechRecognizer = BuildRecognizerFromWaveFile(s);
+            objsToClose.push(r);
+
+            r.recognizeOnceAsync(
```
BrianMouncer, line 1771: >r.recognizeOnceAsync [](http://example.com/codeflow?start=12&length=20)

from experience with what broke in previous SDKs...

I would add another test, that calls recognize twice, in a way that the token should expire, and verify that the header is updated with a second and different guid.

becasue that has caused us issues in the past, where speech works untill the first token expires, and then it didn;t refresh, or there was bad latency in getting the new token.

glharper: This test actually exists in [LongRunning/SpeechRecoAuthTokenRefreshTests.ts](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/master/tests/LongRunning/SpeechRecoAuthTokenRefreshTests.ts). It takes 20 minutes, though, so it is by default disabled in ADO builds. Locally, it passed with this change.

PR #441, opened 2021-09-23
Fix long running timeout issue with endpoints

glharper:
From Friedel:
"run a 1hour recognition.

it will run a translation session, simulating a "bad" connection by delaying the audio randomly for 1-5seconds, then resuming the pushstream.

during that time one can see the following behavior:

  - the audio is correctly processed for n*10minutes + 9minutes:58seconds  (i.e., is almost exactly at a 10 minute boundary)
  - audio is delayed for (e.g.) 5 seconds (i.e., the delay will cross the 10minute boundary).
  - after 1second into that delay, the audio buffer seems to be empty
  - after another 1 seconds, the connection is at a 10min boundary and the service disconnects
  - the client immediately re-connects to the service
  - after another 3 seconds the client resumes audio

from that point on, one can see on the client that audio is "pushed" into the pushtream,

HOWEVER

on the service side we dont see any audio comming in and after 5 minutes the no-audio timeout triggers and an error message is sent on the websocket.

BUT

on the sdk side, as a consequence, the client will start to start sending audio on the existing connection again (note: even though the service indicated a audio timeout error, the client sdk does NOT reconnect but re-use the exisiting connection!)

so, since the service processes audio at 2x realtime, after a total of 10min the 10min audio is processed.
This "hides" the error since doing an overall timecheck, processing seems to be normal, however, during this 5minute window, no processing takes place.
"

### DIFF
--- a/tests/LongRunning/TranslationRecoReconnectTests.ts
+++ b/tests/LongRunning/TranslationRecoReconnectTests.ts
@@ -41,7 +41,6 @@ const BuildSpeechConfig: () => sdk.SpeechTranslationConfig = (): sdk.SpeechTrans
     expect(s).not.toBeUndefined();
     return s;
 };
-/*
 
 // Tests client reconnect after speech timeouts.
 test("Reconnect After timeout", (done: jest.DoneCallback) => {
@@ -196,7 +195,6 @@ test("Reconnect After timeout", (done: jest.DoneCallback) => {
             done.fail(err);
         });
 }, 1000 * 60 * 12);
-*/
 
 test("Test new connection on empty push stream for translator", (done: jest.DoneCallback) => {
     // tslint:disable-next-line:no-console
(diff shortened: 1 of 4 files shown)


trrwilson:
For posterity, could you please add a bit of description about the issue and fix approach? Helps a bunch when going back though! If there are public issues to link, that's always good, too.

glharper:
> For posterity, could you please add a bit of description about the issue and fix approach? Helps a bunch when going back though! If there are public issues to link, that's always good, too.

Added Friedel's description of the issue.

tests/LongRunning/TranslationRecoReconnectTests.ts
```diff
@@ -215,50 +213,32 @@ test("Test new connection on empty push stream for translator", (done: jest.Done
...
-    let shortPause: boolean = false;
+    let disconnected: boolean = false;
+    let reconnected: boolean = false;
+    let reconnectTime: number;
 
     const openPushStream = (): sdk.PushAudioInputStream => {
         // create the push stream we need for the speech sdk.
         const pushStream: sdk.PushAudioInputStream = sdk.AudioInputStream.createPushStream();
-        const chunkSize: number = 1024;
+        const chunkSize: number = 512;
```
trrwilson, line 223: curiosity thing: why reduce the chunk size?

glharper: slows down the intake (and thus the sending of "audio" messages), to stretch out the session to 45 minutes to ensure the connection stays up long enough to disconnect and then have enough audio left over for one recognition post-reconnect.

trrwilson:
For the new test input, is there any way we could approximate by looping a smaller input into a stream or similar? Although it's not truly "huge," the new file is something like three times the size of the src and tests folders *combined* (including existing data) and keeping clones lightweight is always a benefit as long as we can do it.

A longer-term approach might be to have a separate testdata project the way we do for some of the other projections (along with the instructions to grab that on demand).

All not blocking, but probably worthwhile to follow up on later.

glharper:
> For the new test input, is there any way we could approximate by looping a smaller input into a stream or similar? Although it's not truly "huge," the new file is something like three times the size of the src and tests folders _combined_ (including existing data) and keeping clones lightweight is always a benefit as long as we can do it.
> 
> A longer-term approach might be to have a separate testdata project the way we do for some of the other projections (along with the instructions to grab that on demand).
> 
> All not blocking, but probably worthwhile to follow up on later.

Given that the new test is LongRunning (so only run on demand, correct?), do we need to include this file at all? I can add a url comment in Settings linking to a place to grab the file.

trrwilson:
> > For the new test input, is there any way we could approximate by looping a smaller input into a stream or similar? Although it's not truly "huge," the new file is something like three times the size of the src and tests folders _combined_ (including existing data) and keeping clones lightweight is always a benefit as long as we can do it.
> > A longer-term approach might be to have a separate testdata project the way we do for some of the other projections (along with the instructions to grab that on demand).
> > All not blocking, but probably worthwhile to follow up on later.
> 
> Given that the new test is LongRunning (so only run on demand, correct?), do we need to include this file at all? I can add a url comment in Settings linking to a place to grab the file.

That sounds ideal to me if we have a good place to put it.

glharper:
> > > For the new test input, is there any way we could approximate by looping a smaller input into a stream or similar? Although it's not truly "huge," the new file is something like three times the size of the src and tests folders _combined_ (including existing data) and keeping clones lightweight is always a benefit as long as we can do it.
> > > A longer-term approach might be to have a separate testdata project the way we do for some of the other projections (along with the instructions to grab that on demand).
> > > All not blocking, but probably worthwhile to follow up on later.
> > 
> > 
> > Given that the new test is LongRunning (so only run on demand, correct?), do we need to include this file at all? I can add a url comment in Settings linking to a place to grab the file.
> 
> That sounds ideal to me if we have a good place to put it.

For now, I can stash it in a personal repo. I'll talk to Brian about a permanent home.

brandom-msft (approved):

PR #468, opened 2022-01-13
Add protocol to agent as workaround

rhurey:
The long version is that agent-base (What the OCSP check agent is based on) uses the callstack to determine what protocol is being used. When Sentry is being used w/ the library, it asks for the protocol from a call stack that doesn't happen to have the expected modules in it, resulting in http being returned.

Except trying to connect to the Speech Service with http results in an error because https is required.

So explicitly set the protocol to whatever URL is being requested when using an agent.


glharper (approved):

glharper (changes requested):
Re-running tests.

glharper (approved):

dargilco (approved):

PR #479, opened 2022-02-08
Update readme.md

dargilco:
And add some comments on what test settings apply to a few test files. This makes it easy to populate settings in order to only run a single test file as part of development.


yulin-li (approved):
:shipit:

PR #486, opened 2022-02-23
check for privAudio for SAD mute/volume API

### LINKED ISSUE #476: SpeakerAudioDestination -> privAudio undefined for volume get & set
Hello there,
It seems to be an issue in **SpeakerAudioDestination** class, as you can call methods (getters and setters) like **volume** & **mute()**, while **this.privAudio** does not necessarily need to be defined. Anyway, this will obviously result in an uncaught error.

Also, is there any way the community can contribute to this respo? I would like to make few PRs
Thank you.

glharper:
fix #476


yulin-li (approved):
:shipit:

PR #487, opened 2022-02-24
Check if 'detailed' property is missing

samjonescode:
Encountering an error, that 'detailed' is undefined, while developing with the speech-sdk. It's possible for the phraseOutput to be defined while its 'detailed' property is not, leading to errors.


glharper (approved):

glharper:
@samjonescode Thanks for submitting this.

PR #494, opened 2022-03-02
Add audio duration in synthesis result, update word boundary

yulin-li:
- Add audio duration in synthesis result;
- Add punctuation and sentence boundary


package.json
```diff
@@ -107,6 +107,7 @@
     "async-disk-cache": "^2.1.0",
     "bent": "^7.3.12",
     "https-proxy-agent": "^4.0.0",
+    "moment": "^2.29.1",
```
glharper, line 110: From [the moment readme](https://momentjs.com/docs/#/use-it/): We recognize that many existing projects may continue to use Moment, but we would like to discourage Moment from being used in new projects going forward. Instead, we would like to [recommend alternatives](https://momentjs.com/docs/#/-project-status/recommendations/) that are excellent choices for use in modern applications today.

Can we use day.js or some other zero dependency library with a note to drop the dependency once Temporal becomes part of the JS standard?

tests/SpeechSynthesisTests.ts
```diff
@@ -144,7 +143,7 @@ test("testGetVoicesAsyncAuthError", async () => {
     const voicesResult: sdk.SynthesisVoicesResult = await r.getVoicesAsync();
     expect(voicesResult).not.toBeUndefined();
     expect(voicesResult.resultId).not.toBeUndefined();
-    expect(voicesResult.reason).toEqual(ResultReason.Canceled);
+    expect(voicesResult.reason).toEqual(sdk.ResultReason.Canceled);
```
glharper, line 146: Nice catch

package.json
yulin-li: Thanks! I didn't notice this readme, I was thinking moment.js was somehow widely used library which has a C# `TimeSpan` equivalent in JS. I think I'd better just revert to use int64 for the time duration.

src/common.speech/SynthesisTurn.ts
```diff
@@ -239,15 +258,27 @@ export class SynthesisTurn {
...
-    private updateTextOffset(text: string): void {
-        if (this.privTextOffset >= 0) {
+    private updateTextOffset(text: string, type: MetadataType): void {
+        if (type === MetadataType.WordBoundary) {
             this.privTextOffset = this.privRawText.indexOf(text, this.privNextSearchTextIndex);
             if (this.privTextOffset >= 0) {
                 this.privNextSearchTextIndex = this.privTextOffset + text.length;
+                if (this.privIsSSML) {
+                    if (this.privRawText.indexOf("<", this.privTextOffset + 1) > this.privRawText.indexOf(">", this.privTextOffset + 1)
+                        && !(text.length >= 2 && text[0] === "<" && text[text.length - 1] === ">")) {
```
glharper, line 268: Cleaner to extract the second part of this check to a const at the beginning of the function since it's not expensive even if not needed, something like:
const isNotBracket: boolean = !(text.length >= 2 && text[0] === "<" && text[text.length - 1] === ">");

And you can reuse isNotBracket (or whatever name you choose) on line 279 below.

src/sdk/PropertyCollection.ts
```diff
@@ -92,7 +96,7 @@ export class PropertyCollection {
      * @member PropertyCollection.prototype.mergeTo
      * @function
      * @public
-     * @param {PropertyCollection} The collection to merge into.
+     * @param {PropertyCollection}  destinationCollection - The collection to merge into.
```
glharper, line 99: Thanks for this. :-)

package.json
yulin-li: updated

yulin-li:
@glharper thx very much for the detailed review, appreciate

src/common.speech/SynthesisTurn.ts
yulin-li: thx, I moved the two check to functions.

glharper (approved):

Issue #469, opened 2022-01-13
Significant delay when activating language recognition


orgads:
Hi,

When we activate language detection we notice a 3.5s delay between `sessionStarted` and `speechStartDetected`.

Without language detection, the delay is roughly 1s.

The delay origin seems to be in the backend. The SDK sends the media packets immediately.

See [Test project](https://github.com/microsoft/cognitive-services-speech-sdk-js/files/7862223/detect-lang-reaction-time.zip) and logs [with](https://github.com/microsoft/cognitive-services-speech-sdk-js/files/7862225/detect.log) and [without](https://github.com/microsoft/cognitive-services-speech-sdk-js/files/7862226/no-detect.log) detect.

I'm not sure if it should be reported here or somewhere else.

cc @yoshigev @ArielMannes

glharper:
@orgads I've contacted the service team and will follow up here when I hear back.

orgads:
We tried both Once and Continuous. Both worked, but we experienced the same delay for both (that's the second boolean in the function attached).

glharper:
@orgads 
When you write "Once and Continuous", you're referring to the speech recognition mode, correct? 

Language ID, a feature of recognition, has multiple modes, two of which are "Single-shot" and "Continuous" (not connected to the speech recognition mode). JS SDK only supports Single-shot, and Single-shot always has a 2.5-3s delay (for both speech recognition modes).

JS SDK does not support Continuous mode for Language ID (but C#/C++/Python SDKs do support Continuous mode for LID). Continuous mode for LID does not have as long of a delay when recognition starts.

Does that make sense?

orgads:
Ok, thank you.

PR #496, opened 2022-03-03
Glharper/continuous lid

glharper:
#469 adding continuous LID with test for continuous recognition

### DIFF
--- a/src/sdk/AutoDetectSourceLanguageConfig.ts
+++ b/src/sdk/AutoDetectSourceLanguageConfig.ts
@@ -98,10 +98,10 @@ export class AutoDetectSourceLanguageConfig {
     public set continuousMode(isContinuous: boolean) {
         if (isContinuous) {
             this.privProperties.setProperty(PropertyId.SpeechServiceConnection_RecognitionEndpointVersion, "2");
-            this.privProperties.setProperty(PropertyId.SpeechServiceConnection_LanguageIdMode, "DetectContinuous");
+            this.privProperties.setProperty(PropertyId.SpeechServiceConnection_ContinuousLanguageIdPriority, "Latency");
         } else {
             this.privProperties.setProperty(PropertyId.SpeechServiceConnection_RecognitionEndpointVersion, "1");
-            this.privProperties.setProperty(PropertyId.SpeechServiceConnection_LanguageIdMode, "DetectAtAudioStart");
+            this.privProperties.setProperty(PropertyId.SpeechServiceConnection_ContinuousLanguageIdPriority, undefined);
         }
     }
 
--- a/src/sdk/AutoDetectSourceLanguageConfig.ts
+++ b/src/sdk/AutoDetectSourceLanguageConfig.ts
@@ -8,6 +8,8 @@ import {
     PropertyId,
     SourceLanguageConfig,
 } from "./Exports";
+import { LanguageIdMode } from "./LanguageIdMode";
+import { LanguageIdPriority } from "./LanguageIdPriority";
 
 /**
  * Language auto detect configuration.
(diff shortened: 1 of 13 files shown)


src/sdk/AutoDetectSourceLanguageConfig.ts
```diff
@@ -88,4 +88,21 @@ export class AutoDetectSourceLanguageConfig {
...
+     * @summary Sets LID operation to continuous mode
+     */
+    public set continuousMode(isContinuous: boolean) {
+        if (isContinuous) {
+            this.privProperties.setProperty(PropertyId.SpeechServiceConnection_RecognitionEndpointVersion, "2");
+            this.privProperties.setProperty(PropertyId.SpeechServiceConnection_LanguageIdMode, "DetectContinuous");
+        } else {
+            this.privProperties.setProperty(PropertyId.SpeechServiceConnection_RecognitionEndpointVersion, "1");
+            this.privProperties.setProperty(PropertyId.SpeechServiceConnection_LanguageIdMode, "DetectAtAudioStart");
+        }
```
dargilco, line 105: What about setting the priority? "Latency" (the default) and "Accuracy"?

dargilco: Also, don't you want to keep JS property names and meanings the same as in rest of Carbon? Looks like here you defined a new property name that only deals with at-start/continuous.

dargilco (changes requested):

src/sdk/PropertyId.ts
```diff
@@ -245,6 +245,13 @@ export enum PropertyId {
      */
     SpeechServiceConnection_EnableAudioLogging,
 
+    /**
+     * A string value representing the language id mode for language detection.
+     * Allowed values are "Continuous", "Segments", and "AtStart".
+     * Added in version 1.21.0
+     */
+    SpeechServiceConnection_LanguageIdMode,
+
```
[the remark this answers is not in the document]
glharper: Removed language referring to Segments

src/sdk/AutoDetectSourceLanguageConfig.ts
glharper: Priority has now been added. Since this is only continuous LID, only "Latency"/"PrioritizeLatency" or a blank string are allowed. When I add the full matrix of LID possibilities, "Accuracy" will become a possible value.

yulin-li (approved):

glharper:
@katem244 FYI

src/sdk/PropertyId.ts
```diff
@@ -245,6 +245,13 @@ export enum PropertyId {
      */
     SpeechServiceConnection_EnableAudioLogging,
 
+    /**
+     * A string value representing the priority for continuous language detection.
+     * Allowed values are "Latency", and "" (for no continous language detection). "Accuracy" is currently unsupported for Continuous LID.
+     * Added in version 1.21.0
+     */
+    SpeechServiceConnection_ContinuousLanguageIdPriority,
```
dargilco, line 253: Typo in comment ("continous").

dargilco, line 253: Why do you allow setting an empty string? (""). I think you should just say that only "Latency" is supported. 

To set at-start LID, there is another property called "SpeechServiceConnection_SingleLanguageIdPriority" which takes values "latency" or "accuracy". I have a work item to change the name to  SpeechServiceConnection_AtStartLanguageIdPriority (and mark the original property as deprecated). I don't see you have that property in JS. Can you add it? How do you currently switch between latency and accuracy in JS for LID with SR? 
The default is at-start LID with Latency mode.

dargilco: Let's have a quick phone call if you want to discuss this.

dargilco (changes requested):

dargilco (approved):

PR #498, opened 2022-03-09
Convert to using ESLint for static code analysis

glharper:


.eslintrc.js
```diff
@@ -0,0 +1,265 @@
...
+        "@typescript-eslint/no-namespace": "error",
+        "@typescript-eslint/no-parameter-properties": "off",
+        "@typescript-eslint/no-shadow": [
+            "off",
+            {
+                "hoist": "all"
+            }
+        ],
+        "@typescript-eslint/no-unused-expressions": "error",
+        "@typescript-eslint/no-use-before-define": "off",
```
BrianMouncer, line 126: >"@typescript-eslint/no-use-before-define": "off", [](http://example.com/codeflow?start=8&length=49)

why are some of these off?

is there a minimal set we have to have on to be SDL/EO compliant?

.eslintrc.js
```diff
@@ -0,0 +1,265 @@
...
+        "guard-for-in": "error",
+        "id-match": "error",
+        "import/order": "error",
+        "jsdoc/check-alignment": "error",
+        "jsdoc/check-indentation": "error",
+        "max-classes-per-file": [
+            "error",
+            1
+        ],
+        "max-len": "off",
```
BrianMouncer, line 189: >max-len": "off [](http://example.com/codeflow?start=9&length=14)

?

BrianMouncer (approved):
:shipit:

.eslintrc.js
glharper: This is the max line length, I'm unsure whether this was actually enforced by tslint (there were exception comments for it). Unsure of the value of this rule, but willing to turn it on if you feel strongly.

.eslintrc.js
glharper: I think eslint brings over all tslint rules that exist, but if it can't find that rule in our pre-existing tslint.json, it sets it to "off".

src/common.browser/CertChecks.ts
```diff
@@ -311,31 +331,31 @@ export class CertCheckAgent {
...
                 if (error) {
                     reject(error);
                     return;
                 }
 
-                const parsedUri: {[k: string]: any} = parse.default(uri);
-                parsedUri.path = parsedUri.pathname;
-                options = { ...options, ...parsedUri };
+                const url = new URL(uri);
+                options = { ...options, host: url.host, protocol: url.protocol, port: url.port, path: url.pathname};
```
glharper, line 342: @rhurey Thoughts? (This removes the url-parse dependency, which breaks gulp. The url-parse repo now says "use the WHATWG URL API if possible", FWIW). Also, this passes OCSP tests.

PR #504, opened 2022-03-18
feat: added seek & rate to SpeakerAudioDestination

calinalexandru:
Hello,

In my application I need to be able to adjust seek & rate of the Audio element, and I don't believe there's a reason for them not to be here.
Also, I set `privAudio = new Audio()` from the constructor to prevent unnecessary undefined check. 
Went ahead and removed some of those checks, although, more expressions can be simplified in the class.


glharper:
@calinalexandru Thanks for submitting this! The code looks good, I'll run this through some manual tests and give the team and chance to look at it.

src/sdk/Audio/SpeakerAudioDestination.ts
```diff
@@ -52,6 +52,7 @@ export class SpeakerAudioDestination implements IAudioDestination, IPlayer {
         this.privId = audioDestinationId ? audioDestinationId : createNoDashGuid();
         this.privIsPaused = false;
         this.privIsClosed = false;
+        this.privAudio = new Audio();
```
yulin-li, line 55: The `undefined` value of  `privaAudio` means the playback is not supported (e.g. unsupported format, see line 134). This would break some current logic. So pls revert.

yulin-li (commented):
Thank you for your contribution! The new added api looks good to me. Added one comment

src/sdk/Audio/SpeakerAudioDestination.ts
calinalexandru: @yulin-li can you please elaborate your response? From what I can see calling method `format` will re-initialize `privAudio` so it doesn't matter if I also set it constructor. I'm only trying to understand if there's a better approach than re-adding `undefined` checks (following null object pattern). Thanks.

PR #520, opened 2022-04-07
Add Node version info for building to README.md

glharper:


README.md
```diff
@@ -27,7 +27,7 @@ npm install microsoft-cognitiveservices-speech-sdk
 
 This source code for the Cognitive Services Speeck SDK (JavaScript) is available in a public [GitHub repository](https://github.com/Microsoft/cognitive-services-speech-sdk-js). You are not required to go through the build process. We create prebuilt packages tuned for your use-cases. These are updated in regular intervals.
 
-In order to build the Speech SDK, ensure that you have [Git](https://git-scm.com/downloads) and [Node.js](https://nodejs.org/) installed.
+In order to build the Speech SDK, ensure that you have [Git](https://git-scm.com/downloads) and [Node.js](https://nodejs.org/) installed. Version requirement for Node: 12.44.0 or higher (or 14.17.0 or higher for Node 14).
```
BrianMouncer, line 30: Do we need to mention anything about NPM or GULP.
I guess GULP is installed if missing, but if they didn't install NPM with Node.JS then nothing would get installed.

e.g. are there people that use "CorePack" and do not install of NPM???
I noticed that was an option when I did my upgrade to latest version of node.js

BrianMouncer (approved):

README.md
glharper: This is the first I've heard of core pack...npm is an integral part of node, since almost all non-trivial node applications have dependencies.

BrianMouncer: It was my first time hearing about it too.  But when I installed Node.JS it listed both and I chose NPM.

PR #527, opened 2022-04-27
fix #525 - don't use bent.* type in public method signature

### LINKED ISSUE #525: [REG 1.20 -> 1.21] Cannot use on node/ts without esModulesInterop
Hi,

Starting from version 1.21.0, we cannot use this package with typescript because it requires `esModulesInterop`, which we don't use (and it breaks our build).

```
node_modules/microsoft-cognitiveservices-speech-sdk/distrib/lib/src/common.browser/RestMessageAdapter.d.ts:2:8 - error TS1259: Module '"node_modules/@types/bent/index"' can only be default-imported using the 'esModuleInterop' flag

2 import bent from "bent";
         ~~~~

  node_modules/@types/bent/index.d.ts:77:1
    77 export = bent;
       ~~~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.
```

glharper:
fix #525 @orgads

### DIFF
--- a/src/common.browser/RestMessageAdapter.ts
+++ b/src/common.browser/RestMessageAdapter.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-import bent, { BentResponse } from "bent";
+import bent, { BentResponse, RequestBody } from "bent";
 import {
     ArgumentNullError,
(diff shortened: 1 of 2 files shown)


BrianMouncer (approved):
:shipit:

dargilco (approved):

PR #537, opened 2022-05-31
ConversationTranscriber should not connect to Conversation (REST) bac…

glharper:
…kend. All conversation management for transcription should be local


glharper:
Internal ref #4264240

ralph-msft (approved):

PR #568, opened 2022-09-15
fix recognized callback when LUIS response connectionMessage is empty

glharper:
See [this issue](https://github.com/Azure-Samples/cognitive-services-speech-sdk/issues/1644)

### DIFF
--- a/src/common.speech/IntentServiceRecognizer.ts
+++ b/src/common.speech/IntentServiceRecognizer.ts
@@ -185,34 +185,38 @@ export class IntentServiceRecognizer extends ServiceRecognizerBase {
                     if (this.privUmbrellaIntent !== undefined) {
                         addedIntent = this.privUmbrellaIntent;
                     }
-                    const intentId = addedIntent.intentName === undefined ? intentResponse.topScoringIntent.intent : addedIntent.intentName;
-                    let reason = ev.result.reason;
 
-                    if (undefined !== intentId) {
-                        reason = ResultReason.RecognizedIntent;
+                    if (!!addedIntent) {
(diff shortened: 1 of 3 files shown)


trrwilson:
Just for posterity (ease of future glanceable review), could you please rechunk the title and ensure we have a short description? I know, I know, bureaucracy -- but it really does help sometimes if you need to go back and look at things :D

src/common.speech/IntentServiceRecognizer.ts
```diff
@@ -178,41 +178,45 @@ export class IntentServiceRecognizer extends ServiceRecognizerBase {
...
-                let addedIntent: AddedLmIntent = this.privAddedLmIntents[intentResponse.topScoringIntent.intent];
 
-                if (this.privUmbrellaIntent !== undefined) {
-                    addedIntent = this.privUmbrellaIntent;
-                }
-
-                if (null !== intentResponse && addedIntent !== undefined) {
-                    const intentId = addedIntent.intentName === undefined ? intentResponse.topScoringIntent.intent : addedIntent.intentName;
-                    let reason = ev.result.reason;
+                if (null !== intentResponse && !!intentResponse.topScoringIntent && !!intentResponse.topScoringIntent.intent) {
```
trrwilson, line 182: Would it be tough to add a test case for this? "No recognized events with continuous intent" seems like a nicely defined condition we could have a case for, and this logic (which looks right to me) is non-trivial enough that guarding it against future regressions seems smart.

trrwilson (approved):

src/common.speech/IntentServiceRecognizer.ts
glharper: Shouldn't be too hard, I'll get started on that now

PR #585, opened 2022-10-25
CTS regression fix

glharper:


src/common.speech/Transcription/TranscriberRecognizer.ts
```diff
@@ -118,7 +118,7 @@ export class TranscriberRecognizer extends Recognizer {
         const channels = format.channels;
         if (channels === 1 && this.properties.getProperty("f0f5debc-f8c9-4892-ac4b-90a7ab359fd2", "false").toLowerCase() !== "true") {
             throw new Error("Single channel audio configuration for ConversationTranscriber is currently under private preview, please contact diarizationrequest@microsoft.com for more details");
-        } else if (channels !== 8) {
+        } else if (channels !== 8 && channels !== 1) {
```
jhakulin, line 121: Does JS support 8 channels configuration as native SDK?

jhakulin: Would it be a bit cleaner if we do as below?:

if (channels === 1) {
    if (this.properties.getProperty("f0f5debc-f8c9-4892-ac4b-90a7ab359fd2", "false").toLowerCase() !== "true") {
        throw new Error("Single channel audio configuration for ConversationTranscriber is currently under private preview, please contact diarizationrequest@microsoft.com for more details");
    }
} else if (channels !== 8) {
        throw new Error(`Unsupported audio configuration: Detected ${channels}-channel audio`);
}

glharper: > Does JS support 8 channels configuration as native SDK?

It does, for 8-channel wav. I haven't tested with 8-channel mic array, but haven't received any customer feedback for 8-channel mic array either.

PR #586, opened 2022-10-25
Use conventional idioms to access connection and phraseListGrammar fr…

glharper:
…om ConversationTranscriber

### DIFF
--- a/src/sdk/Transcription/ConversationTranscriber.ts
+++ b/src/sdk/Transcription/ConversationTranscriber.ts
@@ -7,11 +7,10 @@ import { Contracts } from "../Contracts";
 import {
     AudioConfig,
     CancellationEventArgs,
-    Connection,
     ConversationTranscriptionEventArgs,
-    PhraseListGrammar,
     PropertyCollection,
     PropertyId,
+    Recognizer,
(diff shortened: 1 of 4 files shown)


src/sdk/Transcription/ConversationTranscriber.ts
```diff
@@ -129,25 +128,21 @@ export class ConversationTranscriber implements ConversationTranscriptionHandler
...
-     * @function
-     * @public
-     * @return {Connection} The Connection instance of the recognizer.
+     * @Internal
+     * Internal data member to support fromRecognizer* pattern methods on other classes.
+     * Do not use externally, object returned will change without warning or notice.
      */
-    public get connection(): Connection {
-        return Connection.fromRecognizer(this.privRecognizer);
+    public get recognizer(): object {
```
rhurey, line 135: >recognizer() [](http://example.com/codeflow?start=15&length=12)

Should this be opaque name wise and just use "InternalData" also?

glharper: good call

rhurey (approved):

Issue #549, opened 2022-07-07
npx npm-force-resolution breaks the install


mwoodsy:
When I run an npm install the preinstall script breaks the installation process. I've narrowed the issue down to the `npx npm-force-resolutions`.  This appears to be a [known issue](https://github.com/rogeriochaves/npm-force-resolutions/issues/33) with that library but the only workaround is to downgrade the version of that library to `0.0.3`. This is the error I experience when trying to do an `npm install microsoft-cognitiveservices-speech-sdk`:

```
npx: installed 6 in 1.89s

/home/mwoodsy/.npm/_npx/26220/lib/node_modules/npm-force-resolutions/out/npm_force_resolutions/core.cljs:61
  (go
  ^
Error: Timeout trying to fetch resolutions from npm
    at switch__2144__auto__ (/home/mwoodsy/.npm/_npx/26220/lib/node_modules/npm-force-resolutions/out/npm_force_resolutions/core.cljs:61:3)
    at /home/mwoodsy/.npm/_npx/26220/lib/node_modules/npm-force-resolutions/out/npm_force_resolutions/core.cljs:61:3
    at npm_force_resolutions$core$find_resolutions_$_state_machine__2145__auto____1 (/home/mwoodsy/.npm/_npx/26220/lib/node_modules/npm-force-resolutions/out/npm_force_resolutions/core.js:476:4)
    at cljs.core.async.impl.ioc-helpers/FN-IDX (/home/mwoodsy/.npm/_npx/26220/lib/node_modules/npm-force-resolutions/out/npm_force_resolutions/core.js:492:85)
    at cljs.core.async.impl.ioc-helpers/run-state-machine (/home/mwoodsy/.npm/_npx/26220/lib/node_modules/npm-force-resolutions/out/cljs/core/async/impl/ioc_helpers.cljs:35:23)
    at cljs.core.async.impl.ioc-helpers/run-state-machine-wrapped (/home/mwoodsy/.npm/_npx/26220/lib/node_modules/npm-force-resolutions/out/cljs/core/async/impl/ioc_helpers.cljs:39:6)
    at fret (/home/mwoodsy/.npm/_npx/26220/lib/node_modules/npm-force-resolutions/out/cljs/core/async.cljs:490:22)
    at take-cb (/home/mwoodsy/.npm/_npx/26220/lib/node_modules/npm-force-resolutions/out/cljs/core/async.cljs:188:63)
[...]

PR #588, opened 2022-11-02
use overrides in package.json instead of force-resolutions

glharper:
This should remove all vulnerabilities listed on npm install/npm run build and all build warnings related thereto.
Internal Ref ID 4404953

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "microsoft-cognitiveservices-speech-sdk",
   "author": "Microsoft Corporation",
   "homepage": "https://docs.microsoft.com/azure/cognitive-services/speech-service/",
-  "version": "1.21.0-alpha.0.1",
+  "version": "1.24.1-alpha.0.1",
   "license": "MIT",
   "description": "Microsoft Cognitive Services Speech SDK for JavaScript",
   "keywords": [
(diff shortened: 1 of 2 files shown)


BrianMouncer (approved):

package.json
```diff
@@ -116,10 +116,16 @@
     "uuid": "^8.3.0",
     "ws": "^7.5.6"
   },
-  "resolutions": {
+  "overrides": {
     "extend": "3.0.2",
```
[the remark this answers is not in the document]
glharper: Overrides allow a bit more granular control, where we can force a version when a dependency is a sub-dependency of a specific library. [This link](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) describes the override capability better than I can.

package.json
```diff
@@ -94,7 +94,7 @@
     "lint": "eslint -c .eslintrc.js --ext .ts src",
     "civersion": "node ci/version.js",
     "prepare": "npm run build",
-    "preinstall": "npm install --package-lock-only --ignore-scripts --no-audit && npx npm-force-resolutions"
+    "preinstall": "npm install --package-lock-only --ignore-scripts --no-audit"
```
anantoghosh, line 97: Please read https://github.com/microsoft/cognitive-services-speech-sdk-js/issues/549#issuecomment-1300262225
Running `npm install` in the `preinstall` script here will run it when a user installs this package messing up their node_modules. This also does not work with yarn or pnpm.
Use a alternate script name like `setup` to allow your developers to run this script.

anantoghosh: @glharper @BrianMouncer

PR #592, opened 2022-11-17
Update LID properties

dargilco:
Remove these (to align with the change in Carbon native):
SpeechServiceConnection_AtStartLanguageIdPriority
SpeechServiceConnection_ContinuousLanguageIdPriority

Add this single one:
SpeechServiceConnection_LanguageIdMode

Change code such that LID configuration is driven by a single "mode". This is already exposed (before my change) as the AutoDetectSourceLanguageConfig.mode property, which can take two enum values LanguageIdMode.AtStart and LanguageIdMode.Continuous. They are reflected as a string in the single property SpeechServiceConnection_LanguageIdMode.

We will only support "PrioritizeLatency" in the JS code. No support for "PrioritizeAccuracy".

Update test to make sure speech.context is set properly when you set AutoDetectSourceLanguageConfig.mode to AtStart, Continuous or don't set it all, in which case it defaults to AtStart.

I wrote sample code to test RecognizeOnce SR + AtStart LID, and StartContinuousRecognition SR + AtStart LID / Continuous LID. Everything works nicely. I tested with multilingual WAV file (multiple utterances, each one at different language). I check that v2 endpoint is automatically set for Continuous LID. I checked the values of speech.context and query URL. There is still work to be done to update the automated tests to do the same. I opened a task for that.


README.md
```diff
@@ -33,6 +33,8 @@ This source code for the Cognitive Services Speeck SDK (JavaScript) is available
 
 In order to build the Speech SDK, ensure that you have [Git](https://git-scm.com/downloads) and [Node.js](https://nodejs.org/) installed. Version requirement for Node: 12.44.0 or higher (or 14.17.0 or higher for Node 14).
 
+__Note (Nov. 17, 2022): we are aware of an issue preventing building the SDK with recent Node.js versions. Until this is fixed, please use Node.js versions v16.18.0 or lower. See [previous releases of Node.js](https://nodejs.org/en/download/releases/)__
```
glharper, line 36: This should be corrected now in master, feel free to clone, install, and build again using node 18.

glharper (approved):
Thanks for simplifying this!

README.md
dargilco: @glharper will do, thanks!

PR #604, opened 2022-12-08
Increase maxHeaderSize to 4296

glharper:
Fix https://github.com/Azure-Samples/cognitive-services-speech-sdk/issues/1774


dargilco (approved):

dargilco:
This is good only as a temporary fix. Need to fix the parser to read chunk sizes and skip them until you reach the "data" chunk. In this particular case, the iOS WAV file has a the Apple "FLLR" chunk that needs to be skipped. See https://stackoverflow.com/questions/6284651/avaudiorecorder-doesnt-write-out-proper-wav-file-header

PR #617, opened 2023-01-10
chore: update UUID lib and corresponding types lib

me4502:
The currently used version of the UUID library used has an issue when run in Jest via the jsdom (browser) environment. Due to this, any tests that interface with this library at all must use hacks to either bump the library from the package or to run the test with node versions of packages. This updates the UUID library to v9 which resolves these issues.

Other changes include major performance improvements and some other changes [noted here](https://github.com/uuidjs/uuid/blob/main/CHANGELOG.md)

v9 does however come with a few breaking changes (changes in supported platforms), that as far as I can tell are acceptable here. The README indicates the project requires Node 12+, which is still supported by this change to UUID. The only part I am unsure about is the drop of support for IE11 and Safari 10, as I can't find any browser support documentation for this library. Given IE11 is considered end of life, I am assuming this is fine but thought I'd mention it here to be sure.

I also noticed that the types library was still on a version much earlier than the UUID library, so I've updated that as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "bent": "^7.3.12",
         "https-proxy-agent": "^4.0.0",
         "simple-lru-cache": "0.0.2",
-        "uuid": "^8.3.0",
+        "uuid": "^9.0.0",
         "ws": "^7.5.6"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


glharper:
@me4502 Thank you for this PR, I will build and test the changes, with the aim of including this in version 1.26.0 (due early spring).

glharper:
@me4502 I've tested this change, and it passes. If you could merge from current master, I'll approve and then merge this PR. Thanks again for submitting it.

me4502:
Just rebased, thanks :)

glharper (approved):

PR #628, opened 2023-02-13
remove got dependency to close vulnerability, use bent in tests

glharper:
Dependabot opened [this PR](https://github.com/microsoft/cognitive-services-speech-sdk-js/pull/627), so I replaced the "got" usage in tests with bent.

### DIFF
--- a/tests/LongRunning/SpeechRecoAuthTokenErrorMessageTests.ts
+++ b/tests/LongRunning/SpeechRecoAuthTokenErrorMessageTests.ts
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
-import got from "got";
+import bent, { BentResponse } from "bent";
 import * as sdk from "../../microsoft.cognitiveservices.speech.sdk";
 import { ConsoleLoggingListener } from "../../src/common.browser/Exports";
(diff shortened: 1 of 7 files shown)


BrianMouncer (approved):

tests/LongRunning/SpeechRecoAuthTokenErrorMessageTests.ts
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
-import got from "got";
+import bent, { BentResponse } from "bent";
```
jhakulin, line 3: Does our samples use got as well and have this vulnerability?

jhakulin (approved):

tests/LongRunning/SpeechRecoAuthTokenErrorMessageTests.ts
glharper: Thankfully no, the browser code uses XMLHTTPRequest to grab an auth token, and the example code in the JS sample README uses the axios library to send REST requests. We used the "got" lib initially bc it allows requests to set the user agent (was needed for OCSP tests). Now that OCSP tests are disabled, we can use bent, which itself has zero dependencies.

PR #636, opened 2023-02-24
add enable miscue property setter to PronunciationAssessmentConfig

### LINKED ISSUE #631: EnableMiscue parameter not working when using javascript Speech SDK
The javascript below uses the SpeechSDK.SpeechRecognizer class to return pronunciation assessment results. The code works and I can change pronunciation configuration parameters and see those changes in the results. For example, I can switch GradingSystem between FivePoint and HundredMark. However, when I set enableMiscue to true (it defaults to false), the results do NOT change. The results do not include error types for omission or insertion. It appears the enableMiscue parameter is not working

```
// Set up recognizer
var speechConfig = SpeechSDK.SpeechConfig.fromSubscription("###", "australiaeast");
    speechConfig.speechRecognitionLanguage = "en-US";

    var audioConfig = SpeechSDK.AudioConfig.fromDefaultMicrophoneInput();

    recognizer = new SpeechSDK.SpeechRecognizer(speechConfig, audioConfig);

// Add pronunciation parameters
    var pronunciationConfig = SpeechSDK.PronunciationAssessmentConfig.fromJSON('{"referenceText":"' + referenceText + '","gradingSystem":"HundredMark","granularity":"Word"}');
    pronunciationConfig.enableMiscue = true; // I have also tried including this in the JSON in the line above
    pronunciationConfig.applyTo(recognizer);

// Get results
    recognizer.recognized = function (s, e) {

    let paResult = SpeechSDK.PronunciationAssessmentResult.fromResult(e.result);
}

// starting and stopping using: recognizer.startContinuousRecognitionAsync() and recognizer.stopContinuousRecognitionAsync()
```

glharper:
Fix #631

### DIFF
--- a/src/sdk/PronunciationAssessmentConfig.ts
+++ b/src/sdk/PronunciationAssessmentConfig.ts
@@ -124,7 +124,21 @@ export class PronunciationAssessmentConfig {
      * @param {boolean} enableMiscue - enable miscue.
      */
     public set enableMiscue(enableMiscue: boolean) {
-        this.properties.setProperty(PropertyId.PronunciationAssessment_EnableMiscue, String(enableMiscue));
+        const enableMiscueString = enableMiscue ? "true" : "false";
+        this.properties.setProperty(PropertyId.PronunciationAssessment_EnableMiscue, enableMiscueString);
+    }
+
+    /**
+     * Gets the boolean enableMiscue property.
+     * Added in version 1.26.0
+     * @member PronunciationAssessmentConfig.prototype.enableMiscue
+     * @function
+     * @public
+     * @return {boolean} enableMiscue - enable miscue.
+     */
+    public get enableMiscue(): boolean {
+        const enableMiscueString = this.properties.getProperty(PropertyId.PronunciationAssessment_EnableMiscue, "false");
+        return (enableMiscueString.toLowerCase() === "true");
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/sdk/PronunciationAssessmentConfig.ts
```diff
@@ -115,6 +115,18 @@ export class PronunciationAssessmentConfig {
         this.privPhonemeAlphabet = phonemeAlphabet;
     }
 
+    /**
+     * Sets the boolean enableMiscue property.
+     * Added in version 1.26.0
+     * @member PronunciationAssessmentConfig.prototype.enableMiscue
+     * @function
+     * @public
+     * @param {boolean} enableMiscue - enable miscue.
```
BrianMouncer, line 124: how important is the type mismatch here for JavaScript.

input param is {bolean} got the passed data and function called take {String}.

BrianMouncer (approved):
looks fine, except for the passing a string to a boolean, then back to sting, only to have a case/if statment to set some jason to true or false.

either use string all the way, or not at all.
Don't switch back and forth???

src/sdk/PronunciationAssessmentConfig.ts
glharper: I understand what you're saying. How important is parity with Carbon here, as Carbon uses that backing property.

PR #649, opened 2023-03-13
remove ocsp code

glharper:


gulpfile.js
```diff
@@ -30,9 +30,6 @@
       .pipe(tsProject())
       .pipe(sourcemaps.write('.'))
       .pipe(gulp.dest('distrib/lib'));
-  }, function () {
-    return gulp.src('./external/**/*')
-      .pipe(gulp.dest('./distrib/lib/external/'));
```
BrianMouncer, line 35: So we hand not other external dependencies, other than the OCSP libs?

BrianMouncer (approved):

gulpfile.js
glharper: Correct.

src/common.browser/CertChecks.ts
```diff
@@ -4,48 +4,14 @@
...
-    nextUpdate: number;
-}
-
-interface tbsResponse {
-    tbsResponseData: {
-        responses: tbsUpdateResponse[];
-    };
-}
 
 export class CertCheckAgent {
```
rhurey, line 16: Is this class still needd at all?

PR #652, opened 2023-03-15
Add ref-doc comments related to service-side audio logging. Fix other (missing) comments.

dargilco:
Add some more details related to audio logging on the service side.

I also noticed that some of the ref-doc comments for methods on the SpeechConfig class do not appear in [the public document](https://learn.microsoft.com/javascript/api/microsoft-cognitiveservices-speech-sdk/speechconfig?view=azure-node-latest). By comparing to other methods which do have published ref-doc comments, it looks like the `@summary` tag is not working. I moved the comments to the top, as this is how other ref-doc comments seem to get published.

### DIFF
--- a/src/sdk/SpeechConfig.ts
+++ b/src/sdk/SpeechConfig.ts
@@ -280,109 +280,112 @@ export abstract class SpeechConfig {
     public abstract get region(): string;
 
     /**
+     * Sets a property value that will be passed to service using the specified channel.
+     * Added in version 1.7.0.
      * @member SpeechConfig.prototype.setServiceProperty
      * @function
      * @public
      * @param {name} The name of the property.
      * @param {value} Value to set.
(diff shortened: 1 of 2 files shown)


glharper (approved):

src/sdk/SpeechConfig.ts
```diff
@@ -280,109 +280,112 @@ export abstract class SpeechConfig {
...
      * @member SpeechConfig.prototype.speechSynthesisVoiceName
      * @function
      * @public
-     * Added in version 1.11.0.
      */
     public abstract set speechSynthesisVoiceName(voice: string);
 
     /**
      * Gets the speech synthesis output format.
+    * Added in version 1.11.0.
```
glharper, line 374: Alignment issue for linter

dargilco: Good catch! Fixed. Thanks!

PR #664, opened 2023-04-13
fix JS tests and v2 service handling post-exception revocation

glharper:
What happened: our testing credentials had been in a grace period, allowed to use v1 endpoints to test. That grace period has ended, meaning some service-behavior dependent tests fail. This PR corrects handling of the new "audio.end" message in Translation and updates expected results from the service based on changed v2 service behavior.

I've temporarily disabled some tests where endpoint behavior seems highly divergent from expected behavior, namely the silence timeout tests where an explicit timeout value is not set.

NOTE: This PR does not assess the value of those service-dependent tests, nor the service-dependent clauses therein.

### DIFF
--- a/tests/TelemetryUtil.ts
+++ b/tests/TelemetryUtil.ts
@@ -24,6 +24,9 @@ export const validateTelemetry: (json: string, numPhrases: number, numHypothesis
         expect(hypothesis.length).toEqual(numHypothesis);
     }
 
+    // For review: V2 service does not appear to be sending a PhraseLatencyMs metric, despite sending a 
+    // translation.phrase message
+    /*
     let foundPhrases: number = 0;
     for (const metric of telemetryMessage.Metrics) {
         if (metric.PhraseLatencyMs !== undefined) {
@@ -35,6 +38,7 @@ export const validateTelemetry: (json: string, numPhrases: number, numHypothesis
     }
 
     expect(foundPhrases).toEqual(numPhrases === 0 ? 0 : 1);
+    */
 
     let foundHypothesis: number = 0;
     for (const metric of telemetryMessage.Metrics) {
--- a/tests/TranslationRecognizerBasicsTests.ts
+++ b/tests/TranslationRecognizerBasicsTests.ts
@@ -172,7 +172,7 @@ describe.each([false])("Service based tests", (forceNodeWebSocket: boolean) => {
                     json.indexOf(sessionId) > 0) {
                     try {
                         expect(hypoCounter).toBeGreaterThanOrEqual(1);
-                        validateTelemetry(json, 2, hypoCounter); // 2 phrases because the extra silence at the end of conversation mode.
+                        validateTelemetry(json, 1, hypoCounter); 
                     } catch (error) {
                         done(error);
                     }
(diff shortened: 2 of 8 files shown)


BrianMouncer (approved):
I would be ok with adding some new event, that existing code just would not know to look for, but existing code would continue to work.... but the make existing code break, is really bad karma...  We should give feedback to the service team, to let them know the impact this event name change had on existing customers.

tests/TelemetryUtil.ts
```diff
@@ -24,6 +24,9 @@ export const validateTelemetry: (json: string, numPhrases: number, numHypothesis
         expect(hypothesis.length).toEqual(numHypothesis);
     }
 
+    // For review: V2 service does not appear to be sending a PhraseLatencyMs metric, despite sending a 
+    // translation.phrase message
```
rhurey, line 28: The service doesn't send the message, the client does.

dargilco (approved):

tests/TranslationRecognizerBasicsTests.ts
```diff
@@ -147,7 +147,8 @@ describe.each([false])("Service based tests", (forceNodeWebSocket: boolean) => {
             ServiceRecognizerBase.telemetryData = undefined;
         });
 
-        test("RecognizeOnceAsync1", (done: jest.DoneCallback) => {
+        // CI test in ADO not receiving recognizing event, disabling for now
```
rhurey, line 150: >// CI test in ADO not receiving recognizing event, disabling for now [](http://example.com/codeflow?start=8&length=68)

I believe we changed the native tests to use a longer phrase when we needed a recognizing event.

tests/TranslationRecognizerBasicsTests.ts
```diff
@@ -307,8 +308,8 @@ describe.each([false])("Service based tests", (forceNodeWebSocket: boolean) => {
 
                 // Translation uses the continuous endpoint for all recos, so the order is
                 // recognized then speech end.
-                expect((LAST_RECORDED_EVENT_ID - 1)).toEqual(eventsMap[SpeechEndDetectedEvent]);
-                expect((LAST_RECORDED_EVENT_ID - 2)).toEqual(eventsMap[Recognized]);
+                // expect((LAST_RECORDED_EVENT_ID - 1)).toEqual(eventsMap[SpeechEndDetectedEvent]);
```
rhurey, line 311: >expect((LAST_RECORDED_EVENT_ID - 1)).toEqual(eventsMap[SpeechEndDetectedEvent]); [](http://example.com/codeflow?start=19&length=80)

We don't get SpeechEnd events anymore?

glharper: We're getting them out of order now, apparently, speech.end before translation.phrase. Another item to inform the service team about.

tests/TelemetryUtil.ts
glharper: This is happening because the speech.end is received before translation.phrase.
On handling speech.end, the code calls shrinkBuffers(), which updates the last offset seen.
Then the translation.phrase message is handled, and the telemetry checks the phrase offset to make sure it's greater than the last seen offset. Because of the out-of-order reception, the phrase offset is less than the last seen offset, so telemetry does not push a PhraseLatencyMs object to Metrics, assuming the phrase offset is not known.
(code [here](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/a79b92f5bd774e03113a196402cedbee82f8cf90/src/common.speech/ServiceTelemetryListener.Internal.ts#L68))

tests/TranslationRecognizerBasicsTests.ts
glharper: Updated test and re-enabled

PR #666, opened 2023-04-17
re-enable telemetry test for v2 service

glharper:


tests/TelemetryUtil.ts
```diff
@@ -24,9 +24,6 @@ export const validateTelemetry: (json: string, numPhrases: number, numHypothesis
         expect(hypothesis.length).toEqual(numHypothesis);
     }
 
-    // For review: V2 service does not appear to be sending a PhraseLatencyMs metric, despite sending a 
-    // translation.phrase message
```
dargilco, line 28: "service does not appear to be sending a PhraseLatencyMs metric"... Is this something we need to talk to the service folks about?

dargilco (approved):

tests/TelemetryUtil.ts
glharper: My comment was misinformed, client sends telemetry messages, not service. This PR fixes the lack of PhraseLatencyMs message in telemetry for continuous recognition.

PR #679, opened 2023-05-18
Added 'LocaleName' in VoiceInfo Class

shivsarthak:
### Description:
This pull request adds support for the 'LocaleName' property in the Azure Cognitive Speech SDK, aligning it with the functionality provided by the web API.

### Motivation:
The web API in the Azure Cognitive Speech service at `/cognitiveservices/voices/list` returns a value for 'LocaleName', which indicates the locale or language of the speech input or output. However, this property was missing in the SDK's class variable, causing a mismatch in functionality between the web API and the SDK.

[Reference Link](https://learn.microsoft.com/en-us/azure/cognitive-services/speech-service/rest-text-to-speech?tabs=streaming#get-a-list-of-voices)


glharper (changes requested):
Please also change the corresponding new VoiceInfo instantiation in sdk/SynthesisVoicesResult.ts to match the new constructor signature.

shivsarthak:

PR #685, opened 2023-06-23
adding meeting transcriber

M-Hietala:


src/common.speech/ConversationTranscriberConnectionFactory.ts
```diff
@@ -0,0 +1,111 @@
...
+            if (!endpoint || endpoint.search(QueryParameterNames.Language) === -1) {
+                queryParams[QueryParameterNames.Language] = language;
+            }
+        }
+
+        if (config.autoDetectSourceLanguages !== undefined) {
+            queryParams[QueryParameterNames.EnableLanguageId] = "true";
+        }
+
+        this.setV2UrlParams(config, queryParams, endpoint);
```
glharper, line 61: setting these URL parameters is unique to V2?

src/common.speech/ServiceRecognizerBase.ts
```diff
@@ -159,13 +174,55 @@ export abstract class ServiceRecognizerBase implements IDisposable {
...
+            speakerDiarization.audioSessionId = this.privDiarizationSessionId;
+            // eslint-disable-next-line no-console
+            console.log("***** ServiceRecognizerBase sessionId: " + this.privDiarizationSessionId);
+            speakerDiarization.audioOffsetMs = 0;
+            phraseDetection.speakerDiarization = speakerDiarization;
+            this.privSpeechContext.setSection("phraseDetection", phraseDetection);
+        }
+    }
+
+    protected setLanguageIdJson(): void {
```
glharper, line 219: Thank you for adding this, logical modularization is always appreciated.

src/common.speech/ServiceRecognizerBase.ts
```diff
@@ -159,13 +174,55 @@ export abstract class ServiceRecognizerBase implements IDisposable {
...
+        if (this.privEnableSpeakerId) {
+            this.privDiarizationSessionId = createNoDashGuid();
+            // this.setSpeakerDiarizationJson(phraseDetection);
+        }
 
-        if (recognizerConfig.autoDetectSourceLanguages !== undefined) {
-            const sourceLanguages: string[] = recognizerConfig.autoDetectSourceLanguages.split(",");
+        this.setLanguageIdJson();
+        this.setOutputDetailLevelJson();
+        this.setSpeakerDiarizationJson();
```
glharper, line 184: This call should also be in the subclass constructor, since it depends on first reading the config value into privEnableSpeakerId. The method definition should probably also go in ConversationTranscriptionServiceRecognizer.

src/common.speech/TranscriptionServiceRecognizer.ts
```diff
@@ -142,9 +166,15 @@ export class TranscriptionServiceRecognizer extends ConversationServiceRecognize
     // Encapsulated for derived service recognizers that need to send additional JSON
     protected async sendTranscriptionStartJSON(connection: IConnection): Promise<void> {
         await this.sendSpeechContext(connection, true);
-        const info: ConversationInfo = this.privTranscriberRecognizer.getConversationInfo();
-        const payload: { [id: string]: any } = this.createSpeechEventPayload(info, "start");
-        await this.sendSpeechEvent(connection, payload);
+        if (this.privTranscriberRecognizer.IsMeetingRecognizer()) {
+            const info: MeetingInfo = this.privTranscriberRecognizer.getMeetingInfo();
```
glharper, line 170: This seems like unnecessary duplication since MeetingInfo and ConversationInfo are structurally identical. It may be a bit of a refactor, so for now is there a way to alias MeetingInfo and rename ConversationProperties to something like GroupProperties (so that it can be used in both MeetingInfo and ConversationInfo)?

src/common.speech/ConversationTranscriberConnectionFactory.ts
M-Hietala: ConversationTranscriber always uses v2, these are the URL params that are not currently sent as json.

glharper (approved):

PR #709, opened 2023-08-07
Some fix on VoiceInfo

hkleungai:
- fix: modify tsconfig so that tsc recognize String.prototype.endsWith()
- fix: remove voicePath from VoiceInfo since it is not in web API.
- refactor: extract and re-use the `IVoiceJson` interface
- feat: Add Neutral to `SynthesisVoiceGender` according to the latest API
- feat: Add missing attributes to IVoiceInfo


hkleungai:

yulin-li (approved):

PR #724, opened 2023-08-28
allow audio worklet node load to be skipped if desired

glharper:
1p issue with blob loading

### DIFF
--- a/src/common.browser/PCMRecorder.ts
+++ b/src/common.browser/PCMRecorder.ts
@@ -111,9 +111,7 @@ export class PcmRecorder implements IRecorder {
                         stream: mediaStream,
                     };
                 })
-                .catch((e: any): void => {
-                    // eslint-disable-next-line no-console
-                    console.log(e);
+                .catch((): void => {
                     attachScriptProcessor();
                 });
         } else {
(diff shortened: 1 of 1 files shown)


src/common.browser/PCMRecorder.ts
```diff
@@ -62,8 +62,9 @@ export class PcmRecorder implements IRecorder {
 
         // https://webaudio.github.io/web-audio-api/#audioworklet
         // Using AudioWorklet to improve audio quality and avoid audio glitches due to blocking the UI thread
+        const skipAudioWorklet = !!this.privSpeechProcessorScript && this.privSpeechProcessorScript.toLowerCase() === "ignore";
 
-        if (!!context.audioWorklet) {
+        if (!!context.audioWorklet && !skipAudioWorklet) {
             if (!this.privSpeechProcessorScript) {
                 const workletScript = `class SP extends AudioWorkletProcessor {
```
BrianMouncer, line 69: If this no longer gets run, is there anything waiting on this to finish, that will now deadlock or timeout and return errors?

BrianMouncer (approved):

src/common.browser/PCMRecorder.ts
glharper: If the addModule is skipped, the deprecated ScriptProcessorNode API is used for audio processing. The 1p requesting this accepts this as a mitigation to bypass the blob URL load in addModule. Followup work will be needed for a fix that doesn't use a deprecated API.

PR #732, opened 2023-09-13
use trusted URL for AudioWorklet

glharper:


src/common.browser/WebsocketMessageAdapter.ts
```diff
@@ -4,6 +4,9 @@
 // Node.JS specific web socket / browser support.
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
+/* eslint-disable @typescript-eslint/no-unsafe-call */
```
BrianMouncer, line 7: This PR is suppose to address some teams security best practices?
So I find it ironic that my biggest red flag are these eslint disable "unsafe blah" directives.

Can you please provide some justification for why these are needed, and why it is safe to do so in this case?

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "target": "es5",
-    "module": "commonjs",
+    "target": "es6",
+    "module": "esnext",
```
BrianMouncer, line 4: is "esnext" stable, or will it change underneath our feet from build to build?

BrianMouncer (approved):
see comments.

dargilco (approved):

rhurey (approved):
:shipit:

tsconfig.json
glharper: Understanding what the tsconfig module option does is non-trivial, but it appears have the module option as esnext should not change the output (while setting the target as esnext would).

src/common.browser/WebsocketMessageAdapter.ts
glharper: I'll look at what I can do to remove these disables, but the switch from es5 to es6 is what triggers the build errors in WebsocketMessageAdapter requiring these directives. es6 is stricter about enforcing eslint rules, apparently.

PR #735, opened 2023-09-22
v2 translation hypothesis, conditionally import audioWorkerUrl method using import.meta

glharper:
- add handling for v2 translation hypothesis messages
- React Native customers can't use import.meta, and customers not using browser microphone functionality don't need to deal with import.meta related errors. This conditionally loads the import.meta code if needed, not at SDK library load time.


glharper:
With this change, the jest.mock additions to our tests may not be needed anymore. Will update PR if that turns out to be the case.

dargilco (approved):

PR #737, opened 2023-09-27
avatar synthesizer implementation

yulin-li:
avatar implementation


yinhew (approved):
I've E2E tested this code on different platforms and browsers, and it's working as expected now.

yinhew:
sign-off

yulin-li:
> Looking at the test results, there are a lot of tests failing that didn't fail for master. @yulin-li could you see why they are failing with this PR?

I checked, one more case `SpeechSynthesisTests.ts-testSpeechSynthesizerEvent` is failing now. I will fix it.

glharper:
> > Looking at the test results, there are a lot of tests failing that didn't fail for master. @yulin-li could you see why they are failing with this PR?
> 
> I checked, one more case `SpeechSynthesisTests.ts-testSpeechSynthesizerEvent` is failing now. I will fix it.

I now see that everything that fails currently also is currently failing in master. Approved.

glharper (approved):

PR #757, opened 2023-11-07
fix trailing space in comment on main file

glharper:


BrianMouncer (approved):
was that the only cause of the npm install failure?

glharper:
> was that the only cause of the npm install failure?

From my node fix PR, yes. There's an outstanding test failure for visemes that Yulin is looking at.

PR #768, opened 2023-11-29
show original error if avatar synthesis failed

yulin-li:
partial effort to fix https://github.com/Azure-Samples/cognitive-services-speech-sdk/issues/2164

### DIFF
--- a/src/sdk/AvatarSynthesizer.ts
+++ b/src/sdk/AvatarSynthesizer.ts
@@ -18,6 +18,7 @@ import {
     AvatarEventArgs,
     PropertyCollection,
     PropertyId,
+    ResultReason,
     SpeechConfig,
     SpeechSynthesisOutputFormat,
     SpeechSynthesisResult,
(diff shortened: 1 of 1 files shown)


yinhew (approved):

src/sdk/AvatarSynthesizer.ts
```diff
@@ -156,6 +165,10 @@ export class AvatarSynthesizer extends Synthesizer {
      * @returns {Promise<void>} The promise of the void result.
      */
     public async stopSpeakingAsync(): Promise<void> {
+        while (this.synthesisRequestQueue.length() > 0) {
```
BrianMouncer, line 168: Is there a time limit here, or do you have to worry about an infinite wait?

BrianMouncer (approved):

src/sdk/AvatarSynthesizer.ts
yulin-li: That's a good point, let me check

yulin-li: This is a corner case, I'd like to merge it first to catch the release.

PR #779, opened 2024-01-02
Remove __proto__ usage in RawWebsocketMessage.ts

iddogino:
Currently this line uses __proto__ to get the objects prototype, which is [deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto), and according to Mozilla's docs "controversial and discouraged"). 

More concretely, this prevents this SDK from running in certain runtimes that do not support `__proto__` - like Deno (see here: https://github.com/denoland/deno/issues/4324). I suggest replacing this code to use the more modern `getPrototypeOf`. With that change, I can confirm this library runs on Deno as well now.

### DIFF
--- a/src/common/RawWebsocketMessage.ts
+++ b/src/common/RawWebsocketMessage.ts
@@ -17,7 +17,7 @@ export class RawWebsocketMessage {
         }
 
         // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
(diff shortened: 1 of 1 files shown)


glharper (approved):

glharper:
@iddogino Thank you so much for submitting this PR!

PR #791, opened 2024-02-01
skip sweden CTS test until service fix is rolled out

glharper:


BrianMouncer (commented):
Same test runs twice right, so this one line fixes both failures?

PR #815, opened 2024-04-14
support g722-16khz-64kbps audio output format

yulin-li:


src/sdk/Audio/AudioOutputFormat.ts
```diff
@@ -519,6 +521,28 @@ export class AudioOutputFormatImpl extends AudioStreamFormatImpl {
...
+            case "amr-wb-16000h":
+                return new AudioOutputFormatImpl(
+                    AudioFormatTag.AMR_WB,
+                    1,
+                    16000,
+                    3052,
+                    2,
+                    16,
+                    speechSynthesisOutputFormatString,
+                    speechSynthesisOutputFormatString,
```
BrianMouncer, line 533: are you sure this second line is correct.
looks like cut and past error, when compared to the other formats that have an actual literal format sting in the second position.

src/sdk/Audio/AudioOutputFormat.ts
```diff
@@ -519,6 +521,28 @@ export class AudioOutputFormatImpl extends AudioStreamFormatImpl {
...
+            case "g722-16khz-64kbps":
+                return new AudioOutputFormatImpl(
+                    AudioFormatTag.G722,
+                    1,
+                    16000,
+                    8000,
+                    2,
+                    16,
+                    speechSynthesisOutputFormatString,
+                    speechSynthesisOutputFormatString,
```
BrianMouncer, line 544: same question/comment here

BrianMouncer (approved):

Issue #848, opened 2024-07-26
[Bug]: ChunkedArrayBufferStream throws RangeError when targetChunkSize is not an integer


rseanhall:
### What happened?

I'm trying to transcribe speech from a WAV file from the npm package of version 1.38.0.

```
    let sourceFile: File;
    // let user pick the file
    const audioConfig = AudioConfig.fromWavFileInput(sourceFile);
    let speechConfig: SpeechConfig;
    // initialize speechConfig
    const conversationTranscriber = new ConversationTranscriber(speechConfig, audioConfig);
    // subscribe to events in conversationTranscriber
    conversationTranscriber.startTranscribingAsync();
```

For this specific WAV file, it causes an error:
[problemfile.zip](https://github.com/user-attachments/files/16394221/problemfile.zip)

```
ERROR RangeError: offset is out of bounds

    at Uint8Array.set (<anonymous>)
    at R.writeStreamChunk (ChunkedArrayBufferStream.ts:40:24)
    at j (FileAudioSource.ts:225:24)
    at Y.onload (FileAudioSource.ts:239:21)
    at w.invoke (zone.js:372:26)
    at Object.onInvoke (ng_zone.ts:423:29)
    at w.invoke (zone.js:371:52)
    at j.runGuarded (zone.js:144:47)
    at FileReader.<anonymous> (zone.js:128:29)
```

After some investigating, the problem seems to be in the `ChunkedArrayBufferStream` constructor:

https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/c12dc569cfffc205d0443345cd4a3bd38db0b126/src/common/ChunkedArrayBufferStream.ts#L12-L16

 It's getting initialized with the `targetChunkSize` of `1102.5`. This causes problems because it is used to setup byte arrays, which can't have fractional items.

### Version

1.36.0 (Latest)

### What browser/platform are you seeing the problem on?

Chrome

### Relevant log output

```shell
logging is not relevant here, the problem is client side
```

PR #849, opened 2024-07-26
Ensure chunk size in ChunkedArrayBufferStream is an integer

rseanhall:
See https://github.com/microsoft/cognitive-services-speech-sdk-js/issues/848


rhurey:
Pipeline is failing because of ADO configuration.

Manually ran against merge commit and passes. Build 149400.

PR #868, opened 2024-11-06
Offset Fixes

rhurey:
While the strongly typed instance properties for the audio offsets had been fixed up to account for multiple turns, the underlying JSON in the result objects wasn't.

This addresses that and add test cases for them.

Fixed and re-enabled the silence detection tests.
Also cleaned up the linter errors in a couple of test files.

### DIFF
--- a/src/common.speech/ConversationServiceRecognizer.ts
+++ b/src/common.speech/ConversationServiceRecognizer.ts
@@ -83,25 +83,23 @@ export class ConversationServiceRecognizer extends ServiceRecognizerBase {
         cancellationReason: CancellationReason,
         errorCode: CancellationErrorCode,
         error: string): void {
-            // Implementing to allow inheritance
-            void sessionId;
-            void requestId;
-            void cancellationReason;
-            void errorCode;
-            void error;
-        }
+        // Implementing to allow inheritance
+        void sessionId;
+        void requestId;
+        void cancellationReason;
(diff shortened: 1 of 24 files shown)


src/common.speech/ConversationServiceRecognizer.ts
```diff
@@ -113,49 +111,44 @@ export class ConversationServiceRecognizer extends ServiceRecognizerBase {
                 EnumTranslation.implTranslateErrorDetails(cancellationErrorCode));
 
         } else {
-            if (!(this.privRequestSession.isSpeechEnded && resultReason === ResultReason.NoMatch && simple.RecognitionStatus !== RecognitionStatus.InitialSilenceTimeout)) {
+            if (simple.RecognitionStatus !== RecognitionStatus.EndOfDictation) {
                 if (this.privRecognizerConfig.parameters.getProperty(OutputFormatPropertyName) === OutputFormat[OutputFormat.Simple]) {
```
pankopon, line 115: Is it verified that `requestWordLevelTimestamps()` does internally set `OutputFormat.Detailed` for the condition here and a couple of other places (ref. https://dev.azure.com/speedme/Carbon/_workitems/edit/14947/)?

pankopon (changes requested):

pankopon (approved):

Issue #896, opened 2025-04-06
[Bug]: Infinite wss connections on non-supported closing codes


gfauchart:
### What happened?

Hi!

On Friday as our application reached its quota of concurrent STT request,
After this, we noticed that our endpoint responsible of STT was making more a lot more than one connection to azure STT
<img width="965" alt="Image" src="https://github.com/user-attachments/assets/ffc4beb8-7feb-4c24-a606-8eb130a269ee" />
(calls in red are wss connections to azure speech service)
this is unusual as this endpoint only makes one call to "recognizeOnceAsync".

I think the problem is [here](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/master/src/common.speech/ServiceRecognizerBase.ts#L176) only closure code 1003, 1007, 1002, 4000 would be considered an error and increment the numConnectionAttempt, leading to the infinite loop of connection.

to reproduce the bug, I've created [this gist](https://gist.github.com/gfauchart/40b23dffa58e7a23130b0fa31354c5ec) which closes the sdk connection with a code 4429 (im assuming that's the code we received but I'm not sure yet).
the sdk would then never resolve and keep making new connection after each 4429

### Version

1.36.0 (Latest)

### What browser/platform are you seeing the problem on?

Node

### Relevant log output

```shell

```

PR #897, opened 2025-04-06
ConnectionClosedEvent to only consider success for closure code 1000

gfauchart:
Hello,

this attempt to fix https://github.com/microsoft/cognitive-services-speech-sdk-js/issues/896
This is assuming that statusCode 1000 is the only success code and any other could be considered an error.

### DIFF
--- a/src/common.speech/ServiceRecognizerBase.ts
+++ b/src/common.speech/ServiceRecognizerBase.ts
@@ -151,10 +151,7 @@ export abstract class ServiceRecognizerBase implements IDisposable {
         this.connectionEvents.attach((connectionEvent: ConnectionEvent): void => {
             if (connectionEvent.name === "ConnectionClosedEvent") {
                 const connectionClosedEvent = connectionEvent as ConnectionClosedEvent;
-                if (connectionClosedEvent.statusCode === 1003 ||
-                    connectionClosedEvent.statusCode === 1007 ||
-                    connectionClosedEvent.statusCode === 1002 ||
-                    connectionClosedEvent.statusCode === 4000 ||
+                if (connectionClosedEvent.statusCode !== 1000 ||
                     this.privRequestSession.numConnectionAttempts > this.privRecognizerConfig.maxRetryCount
                 ) {
                     void this.cancelRecognitionLocal(CancellationReason.Error,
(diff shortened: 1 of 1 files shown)


rhurey:
First, thanks for trying to fix this. A permanent retry loop is definitely not the intent and is a problem.

This fix though will block any retry attempts for a closure that isn't 1000, which means the SDK will stop trying to reconnect if there is a network interruption, the service dies unexpectedly, etc. Those are good reasons to reconnect.

The goal of the error list was to have a (small) defined set of permanently fatal errors that would not be retried, and that block of code aborts the recognition if an error code from that list happens, or the max connection attempts are reached.

The problem I think you're running into is that the code to increase the connection attempt counter only [incremented it if the actual connection setup failed](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/468d0c241cfd03a80650e50fde76585face69bf9/src/common.speech/ServiceRecognizerBase.ts#L978), but didn't if the connection was closed due to an error once websocket upgrade had happened.

I *think* moving the call to onRetryConnection() to the top of the while loop will accomplish the intended behavior.
(I also think adding a test case that forces a 4000 error will show the negative behavior and validate the fix. 4000 is invalid language ID requested.)

rhurey (changes requested):
See the comment above.

gfauchart:
Thanks for getting back to me @rhurey 

I've implemented the suggested change, and it indeed also fixes the issue of infinite loop! (or at least with the mock stt from the gist)
if the STT sends 4429, the sdk would try to connect 5times to the wss server then give up with
```
{"privResultId":"[redacted]","privReason":1,"privErrorDetails":"TooManyRequests websocket error code: 4429","privProperties":{"privKeys":["CancellationErrorCode"],"privValues":["ConnectionFailure"]}}
```

Let me know if this change works for you, or if you have other feedbacks ☺️

gfauchart:
Hi, I saw the pipeline failed,
if this is a problem with the approach feel free to make the fix on a different branch and I will close this PR!
(if you can just tag me on the PR that would be great so I know when the bug is fixed 🙏)

rhurey:
@gfauchart, it's kinda odd. I don't believe it's related to your change, but something is tripping up the script that passed the keys to the test execution. And it didn't repro when I turned up the logging. (See the one passing and one failing pipeline run.)

I'm looking at it, just a few competing things.

rhurey (approved):

rhurey:
@gfauchart, I know what's wrong. It's not you, it's us.

If the new test config system doesn't resolve this, I'll look at how to bypass the error. All the tests are passing.

PR #910, opened 2025-05-10
Introduce redirection capabilities to the fromEndpoint(...) methods

rhurey:
Introduce redirection capabilities to the fromEndpoint(...) methods for SR, S2S, TTS, and CTS.


src/common.speech/ConversationTranscriberConnectionFactory.ts
```diff
@@ -60,6 +63,18 @@ export class ConversationTranscriberConnectionFactory extends ConnectionFactoryB
 
         this.setV2UrlParams(config, queryParams, endpoint);
 
+        if (!!endpoint) {
+            const endpointUrl = new URL(endpoint);
+            const pathName = endpointUrl.pathname;
+
+            if (pathName === "" || pathName === "/") {
```
xitzhang, line 70: Should we consider endpoint like: "wss://{private domain}..../stt/v2" or sth like this? It needs redirection but path is not empty

xitzhang (approved):

src/common.speech/ConversationTranscriberConnectionFactory.ts
rhurey: We can't know what customers have done with containers and where 1P services are hosing their own copies of the front end, so rather than risk being wrong I'm being conservative and taking the approach that "If you provide a host and path, we're going to just accept it and move forward."

xitzhang (approved):

PR #911, opened 2025-05-13
[test]enable aad tests

xitzhang:


rhurey:
How redundant are these to the connection tests that use SpeechConnectionType?

xitzhang:
> How redundant are these to the connection tests that use SpeechConnectionType?

It's a cloud test. I will merge it to SpeechConnectionType later

PR #915, opened 2025-05-21
Fix out of range exception from segmentation parameters

xitzhang:


src/sdk/PropertyId.ts
```diff
@@ -262,6 +262,7 @@ export enum PropertyId {
 
     /**
      * SegmentationMaximumTimeMs represents the maximum length of a spoken phrase when using the Time segmentation strategy.
+     * @member Speech_SegmentationSilenceTimeoutMs must be set in order to use this setting.
```
rhurey, line 265: Doesn't the test below contradict that?

rhurey: I read the test wrong...

rhurey (approved):

PR #925, opened 2025-06-06
Improve inline documentation for segmentation timeout and strategy field

xitzhang:


src/sdk/PropertyId.ts
```diff
@@ -252,11 +252,13 @@ export enum PropertyId {
      * yields results that are too long or too short. Segmentation timeout values that are inappropriately high or low
      * can negatively affect speech-to-text accuracy; this property should be carefully configured and the resulting
      * behavior should be thoroughly validated as intended.
-     *
+     * 
+     * Valid range: **100 to 5,000** milliseconds.
```
rhurey, line 256: We've had some interesting ICM's in the past when documenting static numbers and the service changes the range in the future. Things like "But I used the SDK that documented it as XXX to YYY and not AAA to BBB, so why is it failing?"

Should we doc this on learn and point to there for the ranges?

xitzhang: Yes, I added link of learn of silence here. There's no doc about maximum segment, will add it to doc repo

PR #931, opened 2025-06-24
fix: update wrap Timeout worker script into a Blob object

mattpow:
When using speech SDK in a browser extensions, the Timeout worker is unable to initialize due to the restrictive CORS policies in browser extension contexts. This leads to dictation failing within seconds of calling `startContinuousRecognitionAsync`.

By safely wrapping the worker script in a `Blob` object, the Blob URLs are treaded as same-origin. This is already being used in [PCMRecorder](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/fb04a09bc9b0d08dac7741728863d2524e5db8a1/src/common.browser/PCMRecorder.ts#L85) for its worklet.

This change also includes event listeners for safely unloading the Blob for optimal memory management


mattpow:

PR #983, opened 2026-03-06
feat(tts): add text input streaming support for speech synthesis request

007DXR:
This PR introduces TTS text input streaming in the JS Speech SDK, allowing clients to send text incrementally and receive synthesized audio progressively over the v2 WebSocket endpoint.

**What’s included**
- Adds a new request model for streaming text input:
  - `SpeechSynthesisRequest`
  - `SpeechSynthesisRequestInputType` (with `TextStream` support)
  - `SpeechSynthesisRequestInputStream`
- Adds `SpeechSynthesizer.speakAsync(request, cb, err, stream?)` overload for request-based synthesis.
- Implements adapter-side streaming flow:
  - Sends `synthesis.context` with streaming input metadata (`bidirectionalStreamingMode`).
  - Sends incremental text payloads (`text.piece`).
  - Sends end-of-input signal (`text.end`).
- Supports per-request synthesis properties in streaming mode (pitch/rate/volume/style/temperature/customLexiconUrl/preferLocales).
- Ensures synthesis completion callback ordering is correct by invoking completion event before success callback returns.
- Adds/extends unit + service tests for request creation, stream behavior, progressive streaming events, output streams, and request properties.

**Why**
- Enables LLM/chat-like scenarios where text arrives in chunks and should be synthesized without waiting for full final text.
- Improves end-user latency by allowing progressive audio generation.

**Validation**
- Added request model unit tests (construction, property setters, stream close/write semantics, buffering behavior).
- Added service tests for:
  - streaming with `speakAsync`
  - synthesis lifecycle events
  - file output
  - push/pull audio output streams
  - progressive audio while text is still being sent

**Breaking changes**
- None expected for existing synthesis APIs.

### DIFF
--- a/src/common.speech/SynthesisAdapterBase.ts
+++ b/src/common.speech/SynthesisAdapterBase.ts
@@ -172,8 +172,12 @@ export abstract class SynthesisAdapterBase implements IDisposable {
             this.privSessionAudioDestination.close();
         }
         if (this.privConnectionConfigurationPromise !== undefined) {
-            const connection: IConnection = await this.privConnectionConfigurationPromise;
-            await connection.dispose(reason);
+            try {
+                const connection: IConnection = await this.privConnectionConfigurationPromise;
+                await connection.dispose(reason);
+            } catch {
+                // Connection was never successfully established, nothing to dispose
+            }
         }
     }
 


src/sdk/PropertyId.ts
```diff
@@ -545,4 +545,67 @@ export enum PropertyId {
...
+    /**
+     * The preferred locales for a speech synthesis request.
+     * Added in version 1.37.0
+     * @member PropertyId.SpeechSynthesisRequest_PreferLocales
+     */
+    SpeechSynthesisRequest_PreferLocales,
+
+    /**
+     * The frame timeout interval (in ticks) for speech synthesis.
+     * Added in version 1.37.0
```
yulin-li: these versions are not correct, the next version is 1.49. maybe you can just remove it.

PR #994, opened 2026-03-27
Add Postprocessingoption to js

emilyjiji:


src/sdk/PropertyId.ts
```diff
@@ -313,7 +313,7 @@ export enum PropertyId {
 
     /**
      * A string value specifying which post processing option should be used by service.
-     * Allowed values are "TrueText".
+     * The value is passed directly to the service without modification.
```
rhurey: Should probably leave the TrueText line for now, longer term we'll want to have an aka link to a learn doc that describes the options.

Also, for TrueText, look at what the native code does, JS should probably do the same.

emilyjiji: So, we present it as TrueText for now and then handle it like:
TrueText-> send 5 enrichment values
Everything else -> service accepts/rejects it

PR #1003, opened 2026-04-17
FromOpenRange Support on speechrecognizer

emilyjiji:


src/sdk/SpeechRecognizer.ts
```diff
@@ -64,6 +65,17 @@ export class SpeechRecognizer extends Recognizer {
      */
     public static FromConfig(speechConfig: SpeechConfig, autoDetectSourceLanguageConfig: AutoDetectSourceLanguageConfig, audioConfig?: AudioConfig): SpeechRecognizer {
         const speechConfigImpl: SpeechConfigImpl = speechConfig as SpeechConfigImpl;
+
+        const isOpenRange = autoDetectSourceLanguageConfig.properties.getProperty(
+            PropertyId.SpeechServiceConnection_AutoDetectSourceLanguages, undefined) === AutoDetectSourceLanguagesOpenRangeOptionName;
+        const isMRS = speechConfigImpl.properties.getProperty(
+            PropertyId.SpeechServiceResponse_PostProcessingOption, undefined) === "PostRefinement";
+
+        if (isOpenRange && !isMRS) {
```
rhurey, line 74: This should be an error from the service and not something the client needs to validate.
https://azure.github.io/azure-sdk/general_implementation.html#parameter-validation

Unless the SDK needs to make a decision based on the parameter set, the service should own validating if the scenario is supported or not.

PR #1020, opened 2026-06-19
Add reliable reconnect (multichannel continuation) support

xitzhang:
**Summary**
Implements the multichannel reliable-reconnect protocol for the Speech SDK (JS), bringing it to parity with Carbon. When a connection drops mid-session, the client now resumes the recognition from where the service last acknowledged audio instead of restarting, so no audio (and no recognized text) is lost across a reconnect.

**What it does**
Captures continuation state from service messages: continuation token, per-stream resume offset, and the turn serviceTag.
Rebases offsets: the offset header is turn-relative; it's converted onto the session-absolute timeline and kept monotonic.
Trims the replay buffer to the service-acknowledged offset (mirrors Carbon's OnAcknowledgedAudio → ShrinkReplayBuffer) so a reconnect resends only unacknowledged audio.
Resends continuation on reconnect via the continuation section in speech.context, plus the always-present audio.streams marker.
Activates the server path by sending SeparateChannelProcessing = "true" in speech.config (kept PascalCase to match the case-sensitive server contract) and using a bare universal v2 URL (drops format/language query params so the service returns continuation headers).

**Activation**
Gated entirely behind SPEECH-EnableMultiChannelProcessing. Normal single-stream recognition is unchanged — none of the new headers/sections/flags are emitted unless multichannel is enabled.
Tests
Offline unit tests for the state machine and buffer trimming: [ReliableReconnectTests.ts](vscode-file://vscode-app/c:/Users/xitzhang/AppData/Local/Programs/Microsoft%20VS%20Code/0958016b2a/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [ReplayableAudioNodeTests.ts](vscode-file://vscode-app/c:/Users/xitzhang/AppData/Local/Programs/Microsoft%20VS%20Code/0958016b2a/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (25/25 passing).
Live multichannel reconnect coverage: [SpeechRecoReconnectTests.ts](vscode-file://vscode-app/c:/Users/xitzhang/AppData/Local/Programs/Microsoft%20VS%20Code/0958016b2a/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
npm run lint (src) and tsc --noEmit both clean.

**References**
Reliable reconnect design/protocol: https://speechwiki.azurewebsites.net/partners/protocol-reconnects/


src/common.browser/ReplayableAudioNode.ts
```diff
@@ -99,6 +99,12 @@ export class ReplayableAudioNode implements IAudioStreamNode {
             return;
         }
 
+        // The shrink point must only ever advance; ignore a stale/duplicate offset so the
+        // resume point never moves backwards (matches Carbon's PcmAudioBuffer::DiscardTill).
```
pankopon: Probably should not use "Carbon" in code or PR description.

src/common.speech/ServiceRecognizerBase.ts
```diff
@@ -137,6 +140,11 @@ export abstract class ServiceRecognizerBase implements IDisposable {
         this.privDynamicGrammar = new DynamicGrammarBuilder();
         this.privSpeechContext = new SpeechContext(this.privDynamicGrammar);
         this.privAgentConfig = new AgentConfig();
+        // Multi-channel always relies on the continuation/resume protocol, so reliable reconnect
+        // is enabled implicitly whenever multi-channel processing is active.
+        this.privEnableReliableReconnect =
+            this.privRecognizerConfig.parameters.getProperty("SPEECH-EnableMultiChannelProcessing", "false").toLowerCase() === "true";
```
pankopon, line 146: There is meant to be a PropertyId enum instead of just a string in the public release, could add it JS already.

src/common.speech/ServiceRecognizerBase.ts
```diff
@@ -618,6 +630,12 @@ export abstract class ServiceRecognizerBase implements IDisposable {
             const deviceInfo: ISpeechConfigAudioDevice = await this.audioSource.deviceInfo;
             this.privIsLiveAudio = deviceInfo.type && deviceInfo.type === type.Microphones;
 
+            // SeparateChannelProcessing activates per-channel (multichannel3) processing and the
```
pankopon: Probably better not to refer to "multichannel3" (a pre-release service flight flag and internal feature name).

src/common.browser/ReplayableAudioNode.ts
pankopon: Maybe "native SDK" or something instead.

src/common.speech/ServiceRecognizerBase.ts
pankopon: Or can also update later, looks like this was already noted elsewhere.

xitzhang: yeah, I was just considering service is not GA yet so I add something for internal use for now. I will update a public property id after service finalize the GA

src/common.browser/ReplayableAudioNode.ts
xitzhang: Yes, I removed carbon here

src/common.speech/ServiceRecognizerBase.ts
xitzhang: Yes, removed

src/common.speech/ReconnectContinuationState.ts
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * The stream id used for the single logical audio stream the SDK currently sends.
+     */
+    public get defaultStreamId(): string {
+        return this.privDefaultStreamId;
+    }
+
+    /**
+     * The session-absolute resume offset (100ns ticks) most recently acknowledged by the
+     * service, or undefined if none has arrived. Mirrors Carbon's g_audioContinuationOffset;
```
pankopon: There' still Carbon here and some other places...

src/common.speech/SpeechServiceConfig.ts
```diff
@@ -109,6 +112,9 @@ export interface ISpeechConfigAudioDevice {
     samplerate: number;
     bitspersample: number;
     channelcount: number;
+    // Activates separate channel (multichannel3) processing and the reliable-reconnect contract
```
pankopon: Also multichannel3 here...

xitzhang: yes, just removed all carbon/multichannel3 in all comments

pankopon (approved):

